PR #34, opened 2024-06-27
Adding Setting pane to toggle window switcher on/off

### LINKED ISSUE #22: Option to disable Alt-tab functionality
I like the peek functionality, but for alt+tab I still prefer to use [alt-tab](https://github.com/lwouis/alt-tab-macos) which is more mature and has more features and customization. Please add an option to disable this option and stick with just the Dock peek.

hasansultan92:
This PR fixes #22

### DIFF
--- a/DockDoor.xcodeproj/project.pbxproj
+++ b/DockDoor.xcodeproj/project.pbxproj
@@ -37,6 +37,7 @@
 		BBF6C6242C1B4B4400BF1D40 /* KeybindHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = BBF6C6232C1B4B4400BF1D40 /* KeybindHelper.swift */; };
 		BBFAD73A2C287A9C0092F5F1 /* Sparkle in Frameworks */ = {isa = PBXBuildFile; productRef = BBFAD7392C287A9C0092F5F1 /* Sparkle */; };
 		BBFAD73C2C287C770092F5F1 /* UpdateView.swift in Sources */ = {isa = PBXBuildFile; fileRef = BBFAD73B2C287C770092F5F1 /* UpdateView.swift */; };
+		C119BDF22C2AFA7600E06E18 /* WindowSwitcher.swift in Sources */ = {isa = PBXBuildFile; fileRef = C119BDF12C2AFA7600E06E18 /* WindowSwitcher.swift */; };
 /* End PBXBuildFile section */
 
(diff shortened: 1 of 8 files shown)


DockDoor.xcodeproj/project.pbxproj
```diff
@@ -428,13 +434,13 @@
...
 				ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor;
-				CODE_SIGN_ENTITLEMENTS = DockDoor/DockDoor.entitlements;
+				CODE_SIGN_ENTITLEMENTS = DockDoor/DockDoorDebug.entitlements;
 				"CODE_SIGN_IDENTITY[sdk=macosx*]" = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
 				COMBINE_HIDPI_IMAGES = YES;
 				CURRENT_PROJECT_VERSION = 1.0.5;
 				DEVELOPMENT_ASSET_PATHS = "\"DockDoor/Preview Content\"";
-				DEVELOPMENT_TEAM = 2Q775S63Q3;
+				DEVELOPMENT_TEAM = 8M7864P9XP;
```
ejbills, line 443: Looks like your PR changes the development team. Can you remove this from the commit?

DockDoor/AppDelegate.swift
```diff
@@ -40,7 +41,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         } else {
             self.configureMenuBar()
             dockObserver = DockObserver.shared
-            keybindHelper = KeybindHelper.shared
+            if(Defaults[.showWindowSwitcher]){ // Check if the window switching functionality is enabled or not
```
ejbills, line 44: Parentheses have their place in swift, for a simple conditional like this I would just go with no parentheses. They are not needed.

ejbills: Also, comment isn't needed. I can glance at the if conditional and immediately know what it means, which means you did a good job at naming the variables :)

Readable code > comments

DockDoor/Utilities/Misc Utils.swift
```diff
@@ -13,6 +13,13 @@ func quitApp() {
     NSApplication.shared.terminate(nil)
 }
 
+func restartApplication ()-> Void {
```
ejbills, line 16: Thank you for putting this func here, good idea.

PR #39, opened 2024-06-30
Feat: notify user that the menubar icon can be temporarily show by openning the app

ShlomoCode:
CleanShot 2024-06-30 at 20 13 14@2x

### DIFF
--- a/DockDoor/Views/Settings/SettingsView.swift
+++ b/DockDoor/Views/Settings/SettingsView.swift
@@ -33,8 +33,17 @@ struct SettingsView: View {
             Toggle(isOn: $showMenuBarIcon, label: {
                 Text("Show Menu Bar Icon")
             })
-            .onChange(of: showMenuBarIcon) { _, _ in restartApplication() }
-
+            .onChange(of: showMenuBarIcon) { _, isOn in
+                if !isOn {
+                    let alert = NSAlert()
+                    alert.messageText = "Menu Bar Icon Hidden"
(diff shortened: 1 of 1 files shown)


ejbills:
Hey, thank you. I added in your message and just extended the alert function so it's cleaner. Will be in next version:)

ShlomoCode:
I think people will not pay attention to the information when the title is "Restart required"

In addition, it doesn't seem to me that we need a restart to hide the menubar icon from my testing it works great even without a restart, and only for the presentation you need a restart.

ejbills:
Fixed

ShlomoCode:
It now displays:
CleanShot 2024-06-30 at 20 52 23@2x
But clicking "Cancel" has no effect. Why don't we improve the code so that no restart is needed at all?
I can try to make a PR

ShlomoCode:
Something like that:
```
            Toggle(isOn: $showMenuBarIcon, label: {
                Text("Show Menu Bar Icon")
            })
            .onChange(of: showMenuBarIcon) { _, isOn in
                let delegate = NSApplication.shared.delegate as! AppDelegate
                delegate.updateMenuBarIconStatus()
                
                 if !isOn {
                    MessageUtil.showMessage(
                        title: "Menu Bar Icon Hidden",
                        message: "If you need to access the menu bar icon, launch the app to reveal it for 10 seconds.",
                        completion: { _ in }
                    )
                }
            }
```

ShlomoCode:
https://github.com/ejbills/DockDoor/pull/40

PR #40, opened 2024-06-30
Feat: change showMenuBarIcon setting without restarting the app

ShlomoCode:
For good UX, we should apply setting changes immediately when possible


ShlomoCode:
BTW, this mimics the behavior of all [sindresorhus apps](https://sindresorhus.com/apps)

ejbills (approved):

PR #41, opened 2024-06-30
Fix openScreenRecordingPreferences

ShlomoCode:
The previous address opened the main Security page and not the Screen Recording page

Tested on macOS 14.5, it may be different in older versions, but currently the project runs on macOS 14 only


ejbills (approved):

ShlomoCode:
I believe it’s worth doing a squash to keep the merge clean and avoid a messy commit history :)

PR #45, opened 2024-07-01
Feat: add full-screen toggle button to preview window

ShlomoCode:
Currently we don't support windows in full screen so there is no need for a "toggle", but in the future we will, so I built it as a toggle and not as "enterFullScreen"

SF Symbols doesn't seem to include the exact full screen icon that apple uses, but it's very similar. I checked in alt-tab-macos and it [uses](https://github.com/lwouis/alt-tab-macos/blob/master/src/ui/generic-components/TrafficLightButton.swift#L120-L133) a custom drawing, which I'm not sure if the different license (GPL-3.0 vs MIT) allows us to copy.

Demo:

CleanShot 2024-07-01 at 10 43 56@2x

https://github.com/ejbills/DockDoor/assets/78599753/25654e49-080a-4bad-af49-b6a22109c805

### DIFF
--- a/DockDoor/Utilities/WindowUtil.swift
+++ b/DockDoor/Utilities/WindowUtil.swift
@@ -183,6 +183,19 @@ final class WindowUtil {
         }
     }
     
+    /// Toggles the full-screen state of a window.
+    static func toggleFullScreen(windowInfo: WindowInfo) {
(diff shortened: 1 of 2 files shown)


ejbills (approved):
Great PR, thanks!

PR #53, opened 2024-07-03
feat: Improved clickablity of traffic light buttons in preview window

ShlomoCode:
Allows to click on the edge of the button as well, like in the built-in traffic light of macOS 
Ref: https://stackoverflow.com/a/65101136


ShlomoCode:
It seems to be already resolved at https://github.com/ejbills/DockDoor/commit/e9b1c809fa06278e34d9611ecf7690812932f2e0

Issue #8, opened 2024-06-24
Enhancement suggestion (dock peek): clicking a dock icon should close the pop-up window preview


0n0mat0p03ia:
Currently, when dock peeking, the pop-up preview of an app's window(s) remains on screen even after the user clicks the app's dock icon. Since the user is explicitly summoning the app to the foreground by clicking its dock icon, consider closing the pop-up preview window automatically when the user performs this action. Currently, despite clicking the app's dock icon to summon the app, the user still has to either a) select an item in the pop-up preview to dismiss it, or b) click elsewhere to dismiss it. This feels awkward and inelegant.

ejbills:
I will implement this functionality. However, you do not need to click on anything in the hover window to close it. Simply moving your mouse away from the hover window will hide it automatically.

PR #62, opened 2024-07-04
Feat: close preview when dock icon is clicked

ShlomoCode:
closes https://github.com/ejbills/DockDoor/issues/8

I'm a little concerned about what it means in terms of performance. On the other hand, we already track - and save - every mouse **movement**, so apparently the addition here is no longer significant? Thoughts?

### DIFF
--- a/DockDoor/Utilities/DockObserver.swift
+++ b/DockDoor/Utilities/DockObserver.swift
@@ -50,12 +50,15 @@ final class DockObserver {
         
         func eventTapCallback(proxy: CGEventTapProxy, type: CGEventType, event: CGEvent, refcon: UnsafeMutableRawPointer?) -> Unmanaged<CGEvent>? {
             let observer = Unmanaged<DockObserver>.fromOpaque(refcon!).takeUnretainedValue()
-
+            
             if type == .mouseMoved {
                 let mouseLocation = event.location
                 observer.handleMouseEvent(mouseLocation: mouseLocation)
-            } else if type == .rightMouseDown || type == .otherMouseDown {
-                HoverWindow.shared.hideWindow()
+            } else if type == .rightMouseDown || type == .leftMouseDown || type == .otherMouseDown {
+                let mouseLocation = event.location
+                if observer.isMouseWithinDock(mouseLocation) { // Required to allow clicking the traffic light buttons in the preview
+                    HoverWindow.shared.hideWindow()
+                }
             }
 
             return Unmanaged.passUnretained(event)
@@ -63,6 +66,7 @@ final class DockObserver {
         
         let eventsOfInterest: CGEventMask = (1 << CGEventType.mouseMoved.rawValue) |
                                             (1 << CGEventType.rightMouseDown.rawValue) |
+                                            (1 << CGEventType.leftMouseDown.rawValue) |
                                             (1 << CGEventType.otherMouseDown.rawValue)
         
         eventTap = CGEvent.tapCreate(
@@ -227,6 +231,10 @@ final class DockObserver {
         return nil
     }
     
+    func isMouseWithinDock(_ mouseLocation: CGPoint) -> Bool {
+        return getDockIconAtLocation(mouseLocation) != nil
+    }
+    
     static func screenContainingPoint(_ point: CGPoint) -> NSScreen? {
         let screens = NSScreen.screens
         guard let primaryScreen = screens.first else { return nil }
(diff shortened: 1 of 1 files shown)


ShlomoCode:
I did a non-scientific test - I checked the effect of mouse movements/clicks on the CPU and Energy Impact in XCode Profiler.
Mouse movements seem to have a large effect (worrying!), but clicks have no significant effect:


https://github.com/ejbills/DockDoor/assets/78599753/0ae33cda-ba18-4132-b5ff-45f78503f8a4

ejbills:
I actually meant to close that issue. To get around performance concerns, I made the window get dismissed when a dock icon is clicked by listening to any new didWindowActivate notifications, as long as it is not the current active window. As in, if you click another app, it will get dismissed automatically. The only benefit this serves is dismissing the window for the current active app on dock click, which I'm not sure is necessary. 

In regards to fast mouse movements:
We could make a dead zone around the size and position of the dock, and only start listening once it is in that area. I have thought about implementing that, I don't think it would be too difficult.

ShlomoCode:
At the moment there is a small delay of a week both the window and the preview appear and it is confusing:

https://github.com/ejbills/DockDoor/assets/78599753/f5e67339-e7d1-45d7-99d8-972fb0babfbe

It also doesn't work if you click on the already active app.

In the new method it is completely immediate:

https://github.com/ejbills/DockDoor/assets/78599753/438e28c3-0a10-4058-83a0-fe95f5a5a374

And if we implement a tracking area, I don't think the impact on performance will be significant at all

ShlomoCode:
> The only benefit this serves is dismissing the window for the current active app on dock click, which I'm not sure is necessary.

I think what is important here is that the behavior be consistent, although it is indeed less consumed in such a case, but it may be useful for example in the scenario of floating windows

ejbills:
My 2 concerns are: 

- my getDockIconAtLocation method only returns if the application is running, meaning if they click an app icon in the dock to open an app, I'm not sure this will catch it. The previous window did activate approach would however.
- If we do go this route, I would want to work out the window did activate observers so that we don't have overlapping hiding functionalities. But, we would need to figure out the issue with an app that is closed.

ejbills:
Maybe we can switch the previous observers to another, more scoped observer. Could you play around with `NSWorkspace.didActivateApplicationNotification` replacing with `NSWorkspace.didLaunchApplicationNotification`. In the WindowClosureObservers.swift file? I think that would cover all bases!

ShlomoCode:
Did I understand you correctly that you suggest actually using what I suggested (watch clicks and position check) + watching `didLaunchApplicationNotification` for an application that didn't run (and removing `didActivateApplicationNotification` to avoid overlap)?

1. How is it possible to click on an inactive app, after all, there should not be a preview above an inactive app, and if you hover over an active app and then go to click on an inactive one, already when moving right/left to the icon of the inactive one, even before clicking, the preview will close
2. `didLaunchApplicationNotification` does not cover a case where an app launches itself, for example if an app in the background displays an NSAlert, macos will bring the application to the front, even without a click, in which case we do not want the preview to hide the focused app.

...What is actually the disadvantage of overlapping?

ejbills:
The window can exist over an inactive app if the user initially hovers over an active app and then moves to an inactive app. The window will persist. 

For 2, right it won't activate and therefore won't close the window, so what do you mean? If it's already running and they click it, your code will cover that scenario. If they launch a new app, and it's previously inactive, the observer will catch it. 

Meaning we no longer need to listen to 'application did activate' notifications

> using what I suggested (watch clicks and position check) + watching didLaunchApplicationNotification for an app that didn't run

Correct

ShlomoCode:
> The window can exist over an inactive app if the user initially hovers over an active app and then moves to an inactive app. The window will persist.
> 

I didn't understand... how come the window will persist? It should be detected by tracking the mouse position and close the preview even before any click 🤔

> For 2, right it won't activate and therefore won't close the window, so what do you mean? If it's already running and they click it, your code will cover that scenario. If they launch a new app, and it's previously inactive, the observer will catch it.

True, I was talking about a case where an application that is already open in the background, brings itself to the front, without a user click. My example was an app that displays an `NSAlert`

ejbills:
Play around with it, the hide doesn't kick in until the bounds of the window are left, or you have decently moved away from the last active app icon. There is a deadzone to make it smoother when sliding over the dock.

ShlomoCode:
OK. What about #2?

ShlomoCode:
@ejbills I mean it:

> I was talking about a case where an application that is already open in the background, brings itself to the front, without a user click. My example was an app that displays an `NSAlert`

And it can also be an app that opens with a keyboard shortcut (there are a lot of them)

PR #86, opened 2024-07-07
feat(preview): improve app label shadow

ShlomoCode:
1. Fixed excessive shadow with dark theme
2. Applies the shadow to the app icon as well

CleanShot 2024-07-07 at 00 01 18@2x
CleanShot 2024-07-07 at 00 10 58@2x


ejbills:
The background blur is intended so that it can be read on any background. With a reduced shadow, it is hard to read over other text on screen. I'm happy with the current UI for the label.

ShlomoCode:
What about increasing the shadow to the icon as well, for uniformity?

PR #87, opened 2024-07-07
fix: preview image radius

ShlomoCode:
Before:
When using a dark theme in System Settings, there are white corners around the preview image
CleanShot 2024-07-07 at 05 21 35@2x
CleanShot 2024-07-07 at 05 21 39@2x
After:
CleanShot 2024-07-07 at 05 18 19@2x
CleanShot 2024-07-07 at 05 20 19@2x


ShlomoCode:
I am open to other ways to solve the problem.

ejbills (changes requested):
Please move the radius change to here, rather than making a new modifier. This controls the rounded edge for images.

ejbills (approved):

ShlomoCode:
Each software seems to have a different radius. For example, Slack has a much sharper curve than iTerm, and with the additional radius it is cut:
CleanShot 2024-07-08 at 18 31 45@2x
CleanShot 2024-07-08 at 18 29 24@2x

I'll post a better fix soon.

PR #88, opened 2024-07-07
Introducing the usage of custom keybinds set by user to switch windows

### LINKED ISSUE #11: Ability to change ctrl+tab to another key combination
As the title says, I am use to using command + tab to look at previews as I use alt+tab. This feature would be cool

hasansultan92:
This PR closes #11. Instructions to follow setting a custom user keybind is as follows and can be found in the updated readme as well:

- User selects one of the modifiers presented on the screen
- User presses `Record Keybind` button
- User presses a singular key on the keyboard
- Keybind is now set! 

Please ignore the branch name, the intial idea was to have a dropdown but settled for a segmented picker view instead

The default key bind for the window switcher for DockDoor is now CMD + Tab, replacing the default Mac OS window switcher unless otherwise programmed by the user.


ShlomoCode:
@hasansultan92 I'm curious why you didn't use https://github.com/sindresorhus/KeyboardShortcuts?

hasansultan92:
@ShlomoCode good question! I had a couple of reasons to stay away from this

1. The vast majority of users who are using an application like this want to avoid the default Apple window-switching mechanism and want to use the CMD + Tab key bind. The issues we have been closing on the GitHub and myself included have been in favor of using CMD + Tab for this application. The library you sent above "marks" the default Mac OS shortcuts as unavailable for use and to be remapped and CMD + Tab was one of those key binds we could not reuse/remap.
2. My personal preference when working on projects is to reduce external dependencies. What I have implemented is just intercepting the OS events that are fired from the keyboard and then resuing that information to do what we wanted to do, allowing the user to use a custom key bind to enable the window-switching mechanism. I did not see the need to have an external dependency for something the OS is allowing me by native and also allowing me to remap the default OS key binds to our needs.
3. There was a type mismatch I had to deal with when using the library if I recall correctly. I had an implementation with this library in one of my other branches - the more I think about it that implementation probably does not work anymore. Essentially the OS emits the keyboard events with let's say type x but the library would be using type c for the data types. I was still having to convert them manually to do the operations we were looking into and it did not make sense to continue using the library if I could just create something that worked well with the OS's default types

ShlomoCode:
@ejbills Why does it go in as commit "v1.0.16"? In this way it will be terrible in the future to investigate changes in the git history :(

PR #93, opened 2024-07-08
feat: improve window preview shape

ShlomoCode:
Continuation of https://github.com/ejbills/DockDoor/pull/87

Since not all apps have the same level of window corner rounding, we simply set a transparent background, and display the window as it is instead of interfering with its shape. In addition, we define the gradient overlay with a mask so that it does not color the transparent corners as well.

There is duplication here in defining the gradient but I haven't found a nice way to avoid it. I would love to learn!

### DIFF
--- a/DockDoor/Utilities/WindowUtil.swift
+++ b/DockDoor/Utilities/WindowUtil.swift
@@ -76,9 +76,10 @@ final class WindowUtil {
         
         // Configure the stream to capture only the window content
         config.scalesToFit = false
+        config.backgroundColor = .clear
         config.ignoreGlobalClipDisplay = true
         config.ignoreShadowsDisplay = true
-        config.shouldBeOpaque = true
+        config.shouldBeOpaque = false
         if #available(macOS 14.2, *) { config.includeChildWindows = false }
         config.width = Int(window.frame.width)
         config.height = Int(window.frame.height)
--- a/DockDoor/Views/Hover Window/HoverWindow.swift
+++ b/DockDoor/Views/Hover Window/HoverWindow.swift
@@ -264,6 +264,8 @@ struct HoverView: View {
     let dockPosition: DockPosition
     let bestGuessMonitor: NSScreen
     
+    @Default(.uniformCardRadius) var uniformCardRadius
+    
     @State private var showWindows: Bool = false
     @State private var hasAppeared: Bool = false
     @State private var appIcon: NSImage? = nil
@@ -304,7 +306,7 @@ struct HoverView: View {
                         ForEach(activeWindows.indices, id: \.self) { index in
                             WindowPreview(windowInfo: activeWindows[index], onTap: onWindowTap, index: index,
                                           dockPosition: dockPosition, maxWindowDimension: maxWindowDimension,
-                                          bestGuessMonitor: bestGuessMonitor)
+                                          bestGuessMonitor: bestGuessMonitor, uniformCardRadius: uniformCardRadius)
                                 .id("\(appName)-\(index)")
                         }
                     }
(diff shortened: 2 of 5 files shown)


ejbills:
I renamed a commit based on your feedback and force pushed, seems it broke your PR, sorry about that!

ShlomoCode:
fixed

ejbills:
I would recommend just abstracting it out to a ViewModifier, like so:
```swift
import SwiftUI

struct GradientOverlay: ViewModifier {
    var selected: Bool
    
    func body(content: Content) -> some View {
        content.overlay {
            if selected {
                FluidGradient(
                    blobs: [.purple, .blue, .green, .yellow, .red, .purple].shuffled(),
                    highlights: [.red, .orange, .pink, .blue, .purple].shuffled(),
                    speed: 0.45,
                    blur: 0.75
                ).opacity(0.125)
            }
        }
    }
}
```

And then we don't need to redefine the gradient view twice. Instead, we can now just use: 
`.modifier(GradientOverlay(selected: someBoolInput))`

ejbills:
@ShlomoCode I am pending to release the next patch version, I will hold off until you implement what I mentioned so that no one complains about the window radius. Just ping me once finished!
:)

ShlomoCode:
@ejbills I'm stuck here:
```swift
import SwiftUI
import Defaults
import FluidGradient

struct GradientOverlay: ViewModifier {
    var selected: Bool
    var mask: Image?
    
    func body(content: Content) -> some View {
        content.overlay {
            if selected {
                let gradient = FluidGradient(
                    blobs: [.purple, .blue, .green, .yellow, .red, .purple].shuffled(),
                    highlights: [.red, .orange, .pink, .blue, .purple].shuffled(),
                    speed: 0.45,
                    blur: 0.75
                ).opacity(0.125)
                if mask != nil {
                    gradient.mask(mask)
                }
            }
        }
    }
}

struct WindowPreview: View {
    // ....
    private func windowContent(isMinimized: Bool, isHidden: Bool, isSelected: Bool) -> some View {
        Group {
            if isMinimized || isHidden {
                let width = maxWindowDimension.x > 300 ? maxWindowDimension.x : 300
                let labelText = isMinimized ? "Minimized" : "Hidden"
                
                HStack(spacing: 16) {
                    Image(systemName: "eye.slash.fill")
                        .font(.system(size: 20))
                        .foregroundColor(.secondary)
                    
                    Divider()
                    
                    VStack(alignment: .leading) {
                        Text(labelText)
                            .font(.caption)
                            .foregroundStyle(.secondary)
                        
                        TheMarquee(width: width - 100, secsBeforeLooping: 2, speedPtsPerSec: 30, nonMovingAlignment: .leading) {
                            Text(windowInfo.windowName ?? "\(labelText) window")
                                .font(.system(size: 12, weight: .medium))
                                .foregroundStyle(.primary)
                        }
                    }
                }
                .padding(.horizontal, 8)
                .padding()
                .frame(width: width)
                .frame(height: 60)
                .modifier(GradientOverlay(selected: isSelected))
            } else if let cgImage = windowInfo.image {
                let image = Image(decorative: cgImage, scale: 1.0)
                    .resizable()
                    .aspectRatio(contentMode: .fill)
                
                image.modifier(GradientOverlay(selected: isSelected, mask: image))
            }
        }
        .frame(width: isMinimized || isHidden ? nil : calculatedSize.width,
               height: isMinimized || isHidden ? nil : calculatedSize.height,
               alignment: .center)
        .frame(maxWidth: calculatedMaxDimensions?.width, maxHeight: calculatedMaxDimensions?.height)
    }
}
```
I get a strange error `No exact matches in reference to static method 'buildExpression'`

ejbills:
Check latest PR

ejbills:
But, I honestly think I prefer the uniform-clipped windows. Could you maybe make this an option? Something like "use uniform image preview radius".

ShlomoCode:
`useUniformWindowPreviewRadius`

ShlomoCode:
Why not this way? it's shorter and same effect
```
if Defaults[.useUniformWindowPreviewRadius] {
    image.clipShape(RoundedRectangle(cornerRadius: 10, style: .continuous))
}
image.overlay(isSelected ? fluidGradient().mask(image) : nil)
```

PR #96, opened 2024-07-08
Feat:  settings to customize hover window title style

chrisharper22:
This allows the customization of the Titles for the Previews when hovering over an app in the dock, according to the user's settings. This does not affect the Window Switcher.

Examples:

Embedded:
Embedded Title Style

Popover:
Popover

Note: This is more of a proof of concept. With the current layout of the views, it gets messy quite quickly. Feel free to take a further look at this.

### DIFF
--- a/appcast.xml
+++ b/appcast.xml
@@ -3,12 +3,12 @@
     <channel>
         <title>DockDoor</title>
         <item>
-            <title>1.0.16</title>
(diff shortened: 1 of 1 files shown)


ejbills:
Hey Chris, this is awesome! It looks like there are some merge conflicts, I will take a look once those are fixed.

PR #122, opened 2024-07-09
feat: enforce conventional commits for PRs title

### LINKED ISSUE #85: Enforcement Conventional Commits
Using Conventional Commits makes the change history and intent clearer, especially in the long run. It also allows the creation of a changelog divided into sections (feat, fix, etc.) automatically by GitHub when creating a release.

Refs:
https://www.conventionalcommits.org/en/v1.0.0/
https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13

ShlomoCode:
Enforcing Conventional Commits in the PRs title using https://github.com/amannn/action-semantic-pull-request

The workflow is copied from https://github.com/filebrowser/filebrowser/blob/master/.github/workflows/pr-lint.yaml, you can see this an it in action [here](https://github.com/filebrowser/filebrowser/pull/3235#issuecomment-2119115874).

closes #85

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -1,0 +1,46 @@
+name: "Lint PR"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+      - edited
+      - synchronize
+
+permissions:
+  pull-requests: write
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


ejbills (approved):

ShlomoCode:
In fact it should have been a `chore:`... 😀

ShlomoCode:
@ejbills Just note that this only enforces the PR header, so we have to be careful to use squash merge

PR #128, opened 2024-07-10
feat: added customization options for Window Titles + ...

chrisharper22:
This PR adds customization options for Window Titles: where they are shown and if shown at all.

Second, this PR renames the views for Settings in order to create consistency.

Additionally, this PR addresses some verbiage changes that I believe add clarification to what they do. (Note: on existing string keys, I only updated the English locale to prevent breaking other locales; this should either be updated in all locales or have the direct key(s) changed as well.)

Edit: also, apologizes — this PR contains multiple scopes: feature additions & minor code refactoring.


chrisharper22:
Also, should proper string keys be considered?

PR #137, opened 2024-07-10
refactor(defaults): replace numeric enums with named enums; reword keys and variables

ShlomoCode:


ShlomoCode:
I just think we should adopt a fixed convention about storing options as a number. Does it start with `-1` as in `TitleStyle` or with `0`
What do you think?

ejbills:
Let's keep it consistent with the title options. I would increment everything following never (as 1) up 1. Be sure to set the default value in consts to 2, please!

ShlomoCode:
Yes, I think it should be done in TitleStyle as well :)
It makes sense to me that the selection would be in a consistent order, don't you think so?

ejbills:
Yeah I agree

ShlomoCode:
Should I change `never` to `-1` instead of `0`?

ejbills:
Never: -1
Dimmed on window hover (default) 0

The rest follow suit, be sure to update the consts default key to reflect the changed default key value

ShlomoCode:
@ejbills Redy

ShlomoCode:
@ejbills I'm doing some refactoring in the names of the new settings. Please wait with the next release until I'm done, because then it will be painful to change.

ejbills:
OK just ping me

ShlomoCode:
Still working on it.

ejbills:
All good, take your time. I won't release the next version till you're done.

ShlomoCode:
I will open a new one

PR #138, opened 2024-07-10
fix: update French locale

illavoluntas:
Update the French locale with the new strings provided.

When new commits are pushed on the original repository, it kinds of breaks the local repository I have on my Mac (which I updated through Xcode with Integrate > Pull... after syncing the fork).
I have to start over the modifications. Any idea why?


ejbills:
Xcode git integration sucks, download github desktop and use git via that... Should help your issue

ejbills:
See my latest comment I left on your issue!

PR #141, opened 2024-07-11
refactor(defaults): replace numeric enums with named enums; reword keys and variables

ShlomoCode:
Throws away all numeric enums; Eliminates trying to remember every time "so what does `0` mean and what does `3` mean?". Eliminates awkward code snippets like this:
```
case 1, 2: // Embed + Popover
            Text(appName)
```
```
if  showTitlesOnTiles && ((tileTitlesDisplayMode == 1) || (tileTitlesDisplayMode == 2 && CurrentWindow.shared.showingTabMenu) || (tileTitlesDisplayMode == 3 && !CurrentWindow.shared.showingTabMenu)) {
```
Now it's elegant and clear:
```
case .embedded, .popover:
            Text(appName)
```
```
if  showWindowTitle && ((windowTitleDisplayCondition == .always) || (windowTitleDisplayCondition == .windowSwitcherOnly && CurrentWindow.shared.showingTabMenu) || (windowTitleDisplayCondition == .dockPreviewsOnly && !CurrentWindow.shared.showingTabMenu)) {
```

Collects all the types related to the defaults into the `consts.swift` file as `enums`, and also contains the localized name.

Rewrites the names of the keys and variables (mainly new ones that have not yet been included in a version that has been distributed).

### DIFF
--- a/DockDoor/consts.swift
+++ b/DockDoor/consts.swift
@@ -7,33 +7,106 @@
 
 import Cocoa
 import Defaults
-//import Carbon
+import Foundation
 
 let optimisticScreenSizeWidth = NSScreen.main!.frame.width
 let optimisticScreenSizeHeight = NSScreen.main!.frame.height
 
 let roughHeightCap = optimisticScreenSizeHeight / 3
 let roughWidthCap = optimisticScreenSizeWidth / 3
+
 extension Defaults.Keys {
-    static let sizingMultiplier = Key<CGFloat>("sizingMultiplier") { 3 }
-    static let windowPadding = Key<CGFloat>("windowPadding") { 0 }
-    static let openDelay = Key<CGFloat>("openDelay") { 0 }
-    static let windowTitleAlignment = Key<Bool>("windowTitleAlignment") { true } // true = left side, false = right side alignment
+    static let sizingMultiplier = Key<CGFloat>("sizingMultiplier", default: 3 )
+    static let bufferFromDock = Key<CGFloat>("bufferFromDock", default: 0 )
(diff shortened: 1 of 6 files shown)


ShlomoCode:
@ejbills Can you start reviewing it? I'm still looking for small tweaks but the job is done.

DockDoor/consts.swift
ejbills, line 1: Maybe we move the customization enums under the settings, or at least in their own file? To prevent the consts.swift file from becoming too messy.

ShlomoCode: This is not appropriate for the `settings.swift` file because it deals only with the structure of the settings window.
We could create a new file, but I thought it would make sense to have the definition types in the same file...
What would you call a new file?

ejbills: I meant under the "settings" directory under the views directory, but I'm not sure if it belongs under the views category...

Something like "WindowStylingEnums" or "WindowStylingTypes" perhaps?

ShlomoCode: I don't think it belongs to views
It could be `WindowSettingsEnums`, but I'd just leave it as is. If the file gets too big in the future, we can always move

ejbills: Agreed, worth discussing.

chrisharper22: I would agree these enumerator declarations should be moved elsewhere (and code relating to settings should be refactored in general). 

I originally defined these in their parent structs where they are most applicable, which while is the most popular convention, I don't believe is explicitly stated in any official documentation as far as I know. 

I think the preference keys should be in their own file (more so named other than "conts.swift") , and if we keep these enumerations outside of their associated classes, move them into their own files within a settings folder.

PR #143, opened 2024-07-11
chore: update he localization

ShlomoCode:


ejbills (approved):

DockDoor/Localization/Localizable.xcstrings
```diff
@@ -1279,18 +1332,12 @@
...
+        "he" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Window Titles Appear in"
-          }
-        },
-        "fr" : {
-          "stringUnit" : {
-            "state" : "needs_review",
-            "value" : "Alignement du titre de la fenêtre"
```
ejbills, line 1293: @ShlomoCode looks like this is removing some French translations,  can you revert this and put your translation without the FR removal?

ShlomoCode: I did not touch the French tab. It seems to me that this is something that XCode did automatically

ShlomoCode: It seems that XCode deleted it automatically because I changed the key from "Window Title Alignment" to "Window Title Position" (I also want to add an option for "Top Left" and "Top Right")

As long as we don't have permanent translation keys, any change in the wording in the code will break the translations.
But it's a lot of work to go through and determine a meaningful translation key for each and every string...

PR #144, opened 2024-07-11
fix: update-changelog workflow

ShlomoCode:
See https://github.com/rhysd/changelog-from-release/blob/master/action/action.yml and https://github.com/rhysd/changelog-from-release/blob/master/.github/workflows/post-release.yml


ejbills:
the github-action bot does not have permissions to publish a PR on this repo, so I had to sign the user as myself.

PR #124, opened 2024-07-09
fix: release notes in dark mode

ShlomoCode:
it's just https://github.com/sparkle-project/Sparkle/blob/2.x/Resources/ReleaseNotesColorStyle.css with a little more specific selectors to overwrite jekyll style

closes https://github.com/ejbills/DockDoor/issues/123

PR #147, opened 2024-07-11
fix: invalid formatting in changelog 

ejbills:


ShlomoCode:
This is a legal style and created on purpose. This PR ruined the fix of https://github.com/ejbills/DockDoor/pull/124 and more. Cancel it please

ejbills:
> This is a legal and created on purpose. This PR ruined the fix of #124 and more. Cancel it please

I will revert, do you want to fix the formatting?

ShlomoCode:
What problems do you think there are in the design?
CleanShot 2024-07-11 at 05 43 29@2x

ejbills:
https://github.com/ejbills/DockDoor/blob/main/CHANGELOG.md

The MD is has some html tags on top, I'm not sure why.

ejbills:
IMG_4129

ShlomoCode:
Yes, because GitHub Markdown does not support style tags
There should be a better solution... but it seems to me that for now it's more important that the changes are display to the simple end user properly

ShlomoCode:
I think the right solution is jekyll layouts, but I have no experience with that.

ejbills:
> Yes, because GitHub Markdown does not support style tags There should be a better solution... but it seems to me that for now it's more important that the changes are display to the simple end user properly

Yeah, I agree with that... apologies for breaking the layout haha

PR #150, opened 2024-07-11
feat: add top right window title position option

ShlomoCode:


ShlomoCode:
I didn't add a top-left option because it would overlap with the traffic light buttons.

ejbills (approved):

PR #153, opened 2024-07-11
chore: sync localizations with Crowdin

ejbills:


ejbills:
@ShlomoCode is it expected to be adding half a million lines of code? I selected the "add top 30" option in Crowdin.

PR #156, opened 2024-07-11
chore: sync localizations with Crowdin

ejbills:


ShlomoCode:
Why is the actual commit summary "New Crowdin translations by GitHub Action (#156)"? Maybe because there is only one commit?

PR #178, opened 2024-07-12
feat: suggest users contribute translations

ShlomoCode:


ShlomoCode:
The trick of showing the current language is pretty cool, but I think it would look more consistent without it...
CleanShot 2024-07-12 at 04 54 34@2x
CleanShot 2024-07-12 at 04 57 38@2x

It also gets messy in RTL languages.

Issue #185, opened 2024-07-14
Bug? Window title not permanently displayed


brunbjoern:
(currently on 1.12, issue exists longer)

Altough "Show window title" is set to "always" it only shows the window title if I hover above it.
I want to see as quickly as possible to what window I would like to switch.

![Title Preview](https://github.com/user-attachments/assets/79c486d2-0601-4e5d-b06a-0ffdce9b4d5d)

Or did I understand the meaning of this setting wrong?

Many thanks for putting so much effort in this tool! I really appreciate it!

chrisharper22:
This is currently the intended behavior. But after seeing this, yes, it can be misleading. In the current context, "Always" actually means in both forms, not actually always. I will take a look at this.

chrisharper22:
This is adjusted in #187.

ShlomoCode:
The fix will be available in the next version. @ejbills It can be closed as fixed in https://github.com/ejbills/DockDoor/pull/188

PR #187, opened 2024-07-15
fix: [WindowTitleDisplayCondition] change "Always" behavior

chrisharper22:
Make "Always" actually always show a window title. Additionally, when using Window Switcher-applicable settings, show the app title as the window title for clarity. See #185 for additional information.

### DIFF
--- a/DockDoor/Views/Hover Window/WindowPreview.swift
+++ b/DockDoor/Views/Hover Window/WindowPreview.swift
@@ -119,8 +119,15 @@ struct WindowPreview: View {
                     return .topTrailing
                 }
             }()) {
-                if  showWindowTitle && ((windowTitleDisplayCondition == .always) || (windowTitleDisplayCondition == .windowSwitcherOnly && ScreenCenteredFloatingWindow.shared.windowSwitcherActive) || (windowTitleDisplayCondition == .dockPreviewsOnly && !ScreenCenteredFloatingWindow.shared.windowSwitcherActive)) {
-                    windowTitleOverlay(selected: selected)
+                let shouldShowTitleOverlay = showWindowTitle && (
+                    windowTitleDisplayCondition == .always ||
+                    (windowTitleDisplayCondition == .hoverOnly) ||
+                    (windowTitleDisplayCondition == .windowSwitcherOnly && ScreenCenteredFloatingWindow.shared.windowSwitcherActive) ||
+                    (windowTitleDisplayCondition == .dockPreviewsOnly && !ScreenCenteredFloatingWindow.shared.windowSwitcherActive)
+                )
+
+                if shouldShowTitleOverlay {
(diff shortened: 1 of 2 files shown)


ShlomoCode:
Why not simply change the wording to "both" so that it is clearer? After all, there is already a "Hover Window Title Style" setting that controls the design, and allows you to choose "Hidden", this PR doubles it unnecessarily

ShlomoCode:
I don't think it's right to mix the settings (we already have a "Hover Window Title Style" setting with the "hidden" option).
The correct definition structure in my opinion is as follows:
- Show window titles on:
- - levitation
- - Never
- - Always

- Show window titles for:
- - Dock
- - window changer
- - both of them

- Window title style
- - Default
- - Embedded
- - Popover

And the "Hidden" option will be removed from "Window Title Style".

What do you think?

chrisharper22:
@ShlomoCode the "Hover Window Title" is the application name when hovering Dock icons, which is different than the "Window Title" - the document/tab name hovered over the bottom left/right of the actual window preview itself. 

The mixed usage for Hover Window/Window is confusing to users. I attempted to change these, but they got reverted back to their original strings a few days ago.

If I'm not understanding what you're saying, please let me know.

ShlomoCode:
Haha you are right. It's really very confusing, even I thought you were talking about the app name, we really need to fix the confusion in the settings

ShlomoCode:
@chrisharper22 OK, so we're talking about the window title here and not the name of the application. If so, the design is irrelevant.
But I still think it's worth splitting the definition, similar to what we did with appName.
Something like that:
- Show window titles:
- - Never
- - Always
- - On hovering over the preview
- Show window titles for:
- - Dock
- - Window switcher
- - Both

What do you think?

ShlomoCode:
@chrisharper22 What do you think about it? Can it replace this PR?
I think it sets up the settings better. It also answers the confusion in the settings:

https://github.com/user-attachments/assets/baf74cea-d600-4a08-92c7-03ba3c9f6069

Additionally, the window title will always be visible in the window switcher, even if it is the same as the window name.

chrisharper22:
> @chrisharper22 What do you think about this? Can it replace this PR?
> 
> I think it sets up the settings better. It also answers the confusion in the settings:
> 
> 
> 
> https://github.com/user-attachments/assets/baf74cea-d600-4a08-92c7-03ba3c9f6069
> 
> 
> 
> Additionally, the window title will always be visible in the window switcher, even if it is the same as the window name.

This looks great! I'm fine with closing this PR in favor of yours.

ShlomoCode:
See https://github.com/ejbills/DockDoor/pull/188

chrisharper22:
Closing in favor of #188.

ShlomoCode:
@chrisharper22 Is the case fine now? See the updated video in 188

PR #188, opened 2024-07-15
feat: add option to select window title visibility (whenOveringPreview/alwaysVisible)

ShlomoCode:
Replaces PR https://github.com/ejbills/DockDoor/pull/187.

Additionally:
- Replaces the wording "on hover window" in favor of a clearer wording, "on hover preview", both in the code and in the view
- Corrects the double and misleading wording "window title" in favor of "app name"
- Displays the title of the window for the window changer even if it is the same as the name of the application, because in the window viewer the name of the application is not displayed, so it is not duplicate and unnecessary information
- - I'm uncertain what to do when the user disables "Show app name". Thoughts?

Video:

https://github.com/user-attachments/assets/99c8d196-b6a6-4c5b-abb5-e05aa63ff0cc

### DIFF
--- a/DockDoor/Localizable.xcstrings
+++ b/DockDoor/Localizable.xcstrings
@@ -93,13 +93,13 @@
             "value" : ""
           }
         },
-        "nl" : {
+        "nb" : {
           "stringUnit" : {
             "state" : "new",
             "value" : ""
           }
         },
-        "no" : {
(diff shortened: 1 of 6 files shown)


ShlomoCode:
closes https://github.com/ejbills/DockDoor/issues/185

Issue #57, opened 2024-07-03
Enhancement: App Window Interaction


stevenlis:
Thank you so much for sharing your work with the community. 

I'd like to suggest an improvement to the app's interaction. DockDoor is the kind of app that doesn't need an app window since it just integrates a functionality into the system. The only window it has or needs is a Settings/Preferences window. Many developers like to create a menu icon.

I found an app with an interesting design called [Clop](https://lowtechguys.com/clop/).
![CS-EVshz5sZ@2x](https://github.com/ejbills/DockDoor/assets/33796896/b00732cc-60c8-4fdb-8fea-8d78d687949c)

If the app is not running, launch it, and it will start with the settings window displayed.  You can close the settings window. If you launch the app again, the settings window will reappear, allowing you to adjust settings as needed. To quit the app, simply click the "Quit" button. I find it tedious and unnecessary to go to the menu bar, find the icon, click open settings, and then adjust.

Btw I'm currently experiencing a bug where the menu icon fails to appear with "Show Menu Bar Icon" when I launch the app.

I'm on macOS 14.5 (23F79) with ??? (There appears to be no "About" section in the settings window to display the app version.)

ejbills:
I like this, I just am apprehensive as I assumed users wouldn't want it to run silently in the background. This assumption seems to be incorrect, so I will make this the default functionality if the Show Menu Bar Icon toggle is off. Thanks for your suggestion!

PR #201, opened 2024-07-17
feat: open app settings by relaunch

ShlomoCode:
The opening is even if the menu bar icon is enabled, because sometimes many applications are open in the menu bar and there is no way to get to the DockDoor menu, so it seems logical to me that it would always be possible to open DockDoor again to get to the settings.

closes https://github.com/ejbills/DockDoor/issues/57

*************

It is still not perfect, when the user disables "Show Menu Bar Icon" the window loses focus, and may be lost. I try to focus it back but it only brings it back forward and doesn't focus it, so it can get lost in some scenarios. In any case, the user can always simply reopen the app from the Launchpad and the settings window will open and focused again.

https://github.com/user-attachments/assets/c54792b5-653e-4a3d-9086-bcc71093827f

It happens because of this:
```swift
func removeMenuBar() {
        NSApplication.shared.setActivationPolicy(.prohibited) // hide the dock icon
```

If you have a better way to hide the dock icon even when hiding the menus icon, without losing focus, please let me know!

### DIFF
--- a/DockDoor/AppDelegate.swift
+++ b/DockDoor/AppDelegate.swift
@@ -11,6 +11,16 @@ import Defaults
 import Settings
 import Sparkle
 
+class SettingsWindowControllerDelegate: NSObject, NSWindowDelegate {
+    func windowDidBecomeKey(_ notification: Notification) {
+        NSApp.setActivationPolicy(.regular) // Show dock icon on open settings window
+    }
+    
+    func windowWillClose(_ notification: Notification) {
+        NSApp.setActivationPolicy(.accessory) // Hide dock icon back
+    }
+}
+
 class AppDelegate: NSObject, NSApplicationDelegate {
     private var dockObserver: DockObserver?
     private var appClosureObserver: AppClosureObserver?
@@ -21,7 +31,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     
     // settings
     private var firstTimeWindow: NSWindow?
-    lazy var settingsWindowController = SettingsWindowController(
+    private lazy var settingsWindowController = SettingsWindowController(
         panes: [
             GeneralSettingsViewController(),
             AppearanceSettingsViewController(),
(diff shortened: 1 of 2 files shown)


ShlomoCode:
@hasansultan92 Do you have any thoughts?

ejbills:
Did you look into https://developer.apple.com/documentation/appkit/nsapplication/1428664-presentationoptions?language=objc?

Perhaps something like:
`NSApp.presentationOptions.insert(.autoHideMenuBar)`

But, I am unsure how this behaves.

ShlomoCode:
> Did you look into [developer.apple.com/documentation/appkit/nsapplication/1428664-presentationoptions?language=objc](https://developer.apple.com/documentation/appkit/nsapplication/1428664-presentationoptions?language=objc)?
> 
> Perhaps something like: `NSApp.presentationOptions.insert(.autoHideMenuBar)`
> 

I don't think it's related. From the [options page](https://developer.apple.com/documentation/appkit/nsapplicationpresentationoptions):
>Constants that control the presentation of the app, typically for fullscreen apps such as games or kiosks.

The NSApplicationPresentationHideMenuBar option there, for example, refers to hiding the *entire* menu bar.

ShlomoCode:
According to the documentation [`.accessory`](https://developer.apple.com/documentation/appkit/nsapplication/activationpolicy/accessory) should be enough to hide the Dock icon, but in my observation it didn't work. I will try to investigate it again

hasansultan92:
@ShlomoCode my understanding from the issue -and please correct me if I got it wrong- is that people don't want the application take up a whole "window" when opening up the settings, updates, etc. They want to just go to the menu bar and select from the drop down which view they want to go in?

I went through those pieces of documentation that you and @ejbills have sent, yes we should be able to use something like that to tell the system that this application does not need a full fledged "window" unless, infact that is kind of the logic  we will be using most likely in the future when we implement multiple windows in multiple spaces to determine which application has a "window" and which should not have a window to begin with

ShlomoCode:
@hasansultan92 I'm not sure I understand what you mean. The settings window will not be in full screen in any case, the goal is that there will not be a dock icon when the settings are closed.

> infact that is kind of the logic we will be using most likely in the future when we implement multiple windows in multiple spaces to determine which application has a "window" and which should not have a window to begin with

What do you mean?

hasansultan92:
@ShlomoCode it seems like the way an application has something along the lines of an [activationPolicy](https://developer.apple.com/documentation/appkit/nsrunningapplication/1533103-activationpolicy). The applications that form a full fledged window that we interact with fire/emit a .regular event while like you have mentioned above, some applications fire .prohibited.  We are using that information to determine and filter the applications and windows that are "real" and  not

ShlomoCode:
Yes, it's not a bug in the stable version of DockDoor, it's a bug I encountered as soon as I turned off the menu bar icon, the Dock icon immediately popped up

ShlomoCode:
Currently:
- Shows the Dock icon when the Settings window itself is open
- Shows the bar symbol according to the settings
- Not losing focus at all

@ejbills I think it's ready

Issue #243, opened 2024-08-06
[BUG] DockDoor previews are unstable


illavoluntas:
DockDoor randomly does not show the previews when the Dock is hovered.

## Steps to reproduce
- Try to hover the preview to show a large preview, then go away. Try again and randomly no Dock preview nor large preview will appear.
- Restarting the app will fix the problem.

## Environment
 - macOS version: macOS Sonoma 14.5
 - App version: 1.1.5
 
## Note
~~This might be a cause/consequence of #244 because it seems to happen only with Adobe products.~~

gkrawiec:
i might have this same problem, though I cannot link it to first hovering on a preview. It seems to just stop working randomly, but not crashing. Restarting the app fixes it.

illavoluntas:
Author of #230 might have more informations about it.

nkapila6:
I'm facing the same issue. Works fine on a restart and then it just stops randomly.

mchewy:
I have the same issue but is not random for me - happens when the machine wakes from sleep. A quick restart of the dockdoor app fixes it, but would be great if we could get the issue addressed.

illavoluntas:
So this bug seems to be general. It might be a bug in some new features brought by 1.1.5. 
Could @ejbills or @ShlomoCode or anyone else in the development team try to investigate about it?

illavoluntas:
> I have the same issue but is not random for me - happens when the machine wakes from sleep. A quick restart of the dockdoor app fixes it, but would be great if we could get the issue addressed.

You may have spotted the problem. Now I notice DockDoor doesn’t work anymore after sleep.

ShlomoCode:
@mchewy @illavoluntas For me it continues to work even after sleep ("Sleep" item from the apple menu). Can you detail how to reproduce the bug?

ShlomoCode:
@gkrawiec It sounds more like Log Out than Sleep
Make sure that "Launch at login" is enabled for you in the DockDoor settings
![CleanShot 2024-08-13 at 18 04 59@2x](https://github.com/user-attachments/assets/a7bf9ae9-fcd9-4395-b0c4-71246609be98)

gkrawiec:
it happens randomly  while in my account and for sure when I return from another profile after LOCK SCREEN. There is no logout and the app is still open.

ShlomoCode:
[...]

Issue #212, opened 2024-07-18
Feature Request: Fade away and Delay settings for Hover Preview


jacobcassidy:
I was thinking that a great feature would be if the preview had a setting to delay how long it stays around before disappearing when leaving the dock, or even a fade-out effect that can still be activated by hovering over it before it fades out. 

This will be helpful for those who use auto hide/show for the dock so they don't have to make the preview edge right up to the dock to be able to select the hover previews

ejbills:
Will look into this!

Issue #190, opened 2024-07-15
Listening to all mouse movements is a performance concern


ShlomoCode:
> In regards to fast mouse movements:
We could make a dead zone around the size and position of the dock, and only start listening once it is in that area. I have thought about implementing that, I don't think it would be too difficult.

_Originally posted by @ejbills in https://github.com/ejbills/DockDoor/issues/62#issuecomment-2207710289_

@ejbills Can you elaborate on that a bit more?

ejbills:
Using the Dock Util methods, we can approximate an area and location of the dock. Using this, we can immediately rule out wasted icon frame computations based on the mouse being within our approximated dock area.

ShlomoCode:
Oh, I thought macOS had an option to only listen for movements on a certain element or something, and I didn't understand why I can't find it... 😆

ejbills:
Me too, which is why I used that verbiage in my original comment. Is it really not possible? Either way, whether we can make the app only ‘track’ the mouse in that zone or use it to prevent useless dock frame calculations, it should hopefully result in much less CPU usage when the mouse is moved quickly.

ShlomoCode:
@ejbills I found 2 huge breakthroughs, stay tuned!

ejbills:
> @ejbills I found 2 huge breakthroughs, stay tuned!

You have my interest!

ejbills:
@ShlomoCode 

> I hope there won't be too many conflicts with the refactor I'm cooking 😆

We should probably discuss! Mind sharing more details on what you're cooking up?

ShlomoCode:
I researched the Dock's accessibility information and found 2 things:
- We can find the application under the mouse using [`kAXSelectedChildrenAttribute`](https://developer.apple.com/documentation/applicationservices/kaxselectedchildrenattribute), instead of our own calculation based of the mouse position.
As a side effect this solve https://github.com/ejbills/DockDoor/issues/19 as we are now using macOS's determination of what is considered "selected".
- It is possible to reach the bundle URL of a certain dock item element using [`kAXURLAttribute`](https://developer.apple.com/documentation/applicationservices/kaxurlattribute). From there you can get the budle identifier and use it to find the application windows in a completely accurate way according to the budle identifier, instead of the current heuristic (although we will have to implement a fallback for some particular applications that do not have a budle identifier, I know about scrcpy and the Android Studio emulator)

[...]

Issue #54, opened 2024-07-03
Feature Request: Delay to preview next app


brunbjoern:
We have a hover-windows-open-delay setting which delays the time until a preview is shown.
All following previews of other apps will be displayed immediately which sometimes make it difficult to click on an preview to switch to that window.
E.g. my Finder Icon with quite a number of open windows is right next to my Mail icon.
When trying to move the mouse to the Finder window I would like to switch to, a lot of times dockdock switches to the Mail window previews.

This could be prevented by having a configurable delay before displaying the next App's preview windows, e.g. a "Next Window Open Delay".

(I hope I there is not already an existing Issue for this)

brunbjoern:
![Why this is important](https://github.com/user-attachments/assets/b8455de2-83c2-4e10-a93b-09bc1cd73fd6)

I've tried to visualize why a configurable delay to display the next applications dock preview comes quite handy:
If I want to switch to the right window going the direct path (red arrow) with the cursor results in the next application's window pop up (in the sample above: Mail), preventing to ever get to choose the finder window.
Only by not crossing the next apps icon (e.g. going via the green arrow) I am able to switch to the preferred window.

If you add a configurable delay before switching to the next app's dock preview this can be prevented. This is something the mother of all Dock Preview Tools, HyperDock, does to make the usage more intuitive.

brunbjoern:
With delay (or some other magic - it just feels more natural):
https://github.com/user-attachments/assets/05561ac6-674d-4261-9d84-3fd1fafb68da

DockDoor:
https://github.com/user-attachments/assets/349abce7-1d5c-442a-a25c-feee21dc7370

brunbjoern:
With additional delay (or some other magic - whatever, it feels more natural):
https://github.com/user-attachments/assets/648315ab-e63f-4258-a076-84a997a9e655

Current implementation:
https://github.com/user-attachments/assets/5a86adac-065d-4d61-9ab0-58e3546b399c

salvatoreromeo:
I experience the same problem and this delay option would be very useful

ejbills:
https://github.com/user-attachments/assets/2d6eb3a0-caba-4d18-8979-b8dd0c8322db

Check it out, will be in the next version (1.2 is coming soon).

ejbills:
> I just tried the last version 1.1.6 and this feature does not work as expected: the first preview opens after the hover delay, but when I hover on the second app it shows instantly (it does not respect the delay) 

[...]

PR #254, opened 2024-08-14
feat: efficient current Dock item detection and improved window management

ShlomoCode:
## Describe your changes

- Get current dock item element by macOS Dock's detection (`kAXSelectedChildrenChangedNotification` and `kAXSelectedChildrenAttribute`) instead of calculating mouse position.
- Get rid of listening for mouse events, reducing CPU usage to 0% when not interacting\hovering with the Dock.
- Store windows in cache (SpaceWindowCacheManager.swift) by PID instead of BundleID, to deal with apps like scrcpy that not having a BundleID.
- Adds additional logic for better window state management. 
- Addresses window switcher update inconsistencies. 
- Introduces window fade out animations and configuration options. 

Ported from https://github.com/ejbills/DockDoor/pull/214.

closes https://github.com/ejbills/DockDoor/issues/190
closes https://github.com/ejbills/DockDoor/issues/19
closes https://github.com/ejbills/DockDoor/issues/243
closes https://github.com/ejbills/DockDoor/issues/30
closes https://github.com/ejbills/DockDoor/issues/54
closes https://github.com/ejbills/DockDoor/issues/212
closes https://github.com/ejbills/DockDoor/issues/277

## Core Functionality Changes

Fixes some bugs, and significantly improves performance.


ShlomoCode:
The Dock functionality works, the window changer is still problematic.
Main points remaining:
1. https://github.com/ejbills/DockDoor/pull/254/files#diff-017b6b046f3844ad59afc9f7acd992ad75e9b77ec27f24b4186d13130fc73f20R85
2. The windows switcher is very slow to open
3. Maybe we should get the windows using kAXWindowsAttribute

Also, the code needs polishing, and better documentation of the changes in PR.

DockDoor/Utilities/KeybindHelper.swift
```diff
@@ -110,18 +110,14 @@ class KeybindHelper {
 
     private func showHoverWindow() {
         Task { [weak self] in
-            do {
+            guard let self else { return }
+            await MainActor.run { [weak self] in
                 guard let self else { return }
-                let windows = try await WindowUtil.activeWindows(for: "")
```
ejbills, line 115: The activeWindows call to the empty string also serves to update the space cache, I don't think we need to change this part.

DockDoor/Utilities/SpaceWindowCacheManager.swift
ejbills, line 1: What all exactly changes here? Are you removing all the bundle id helper functions as you are instead storing the process identifier?

ShlomoCode: Yes, to deal with Scrcpy not having a bundle id

DockDoor/Utilities/KeybindHelper.swift
ShlomoCode: What cache update do you mean?

ShlomoCode:
@ejbills I'd appreciate help with https://github.com/ejbills/DockDoor/pull/254/files#diff-017b6b046f3844ad59afc9f7acd992ad75e9b77ec27f24b4186d13130fc73f20R85

ejbills:
> @ejbills I'd appreciate help with https://github.com/ejbills/DockDoor/pull/254/files#diff-017b6b046f3844ad59afc9f7acd992ad75e9b77ec27f24b4186d13130fc73f20R85

Took a look, I think we can just store the previous app name string and only allow the hide action when it differs. I believe that is what happened previously.

ShlomoCode:
You can push. You'll probably want to use DockObserver's `lastAppUnderMouse`.

ShlomoCode:
Looks better now, But it seems like quick mouse movement doesn't trigger SwiftUI's `.onHover` :thinking:

DockDoor/Utilities/KeybindHelper.swift
ejbills: For tracking windows in all spaces, the call to activeWindows is needed here.

ejbills:
> You can push. You'll probably want to use DockObserver's `lastAppUnderMouse`.

I just sat down to work on this, did you already do it? The link goes to a different part of the code now, I can't find the TODO.

ejbills:
> I push some changes, Looks better now, But it seems like quick mouse movement doesn't trigger SwiftUI's `.onHover` 🤔

I think it's all working on my end, what isn't working for you?

ShlomoCode:
> > You can push. You'll probably want to use DockObserver's `lastAppUnderMouse`.
> 
> I just sat down to work on this, did you already do it? The link goes to a different part of the code now, I can't find the TODO.

Yes i did it

ShlomoCode:
> > I push some changes, Looks better now, But it seems like quick mouse movement doesn't trigger SwiftUI's `.onHover` 🤔
> 
> I think it's all working on my end, what isn't working for you?

1. Tap on an app icon
2. A preview will appear
3. **Quickly** move the mouse cursor up through the preview
4. The preview is not hidden. If you move the mouse slowly, the preview does close properly

https://github.com/user-attachments/assets/d081bda1-331d-497d-8a3e-5c6c8fcd0513

DockDoor/Utilities/KeybindHelper.swift
ShlomoCode: Didn't calling `activeWindows` with an empty string simply return all that was in the cache?
https://github.com/ejbills/DockDoor/blob/31f8c1b2c8f478e9f85d1dea692f98db715e56b9/DockDoor/Utilities/WindowUtil.swift#L409-L411

ejbills: It did indeed return `getAllWindows`, but not till after it had made a few calls to `fetchWindowInfo` which were called before the return (even with an empty string). And the `fetchWindowInfo` func served to also update the cache [here.](https://github.com/ejbills/DockDoor/blob/31f8c1b2c8f478e9f85d1dea692f98db715e56b9/DockDoor/Utilities/WindowUtil.swift#L473)

ejbills:
> > > I push some changes, Looks better now, But it seems like quick mouse movement doesn't trigger SwiftUI's `.onHover` 🤔
> > 
> > 
> > I think it's all working on my end, what isn't working for you?
> 
>     1. Tap on an app icon
> 
>     2. A preview will appear
> 
>     3. **Quickly** move the mouse cursor up through the preview
> 
>     4. The preview is not hidden. If you move the mouse slowly, the preview does close properly
> 
> 
> CleanShot.2024-08-22.at.06.58.25.mp4

I can't get this to happen, but if I use it while I have an app fullscreen and I have the full size window hover action, the app seems to freak out. Can you reproduce?

DockDoor/Utilities/KeybindHelper.swift
ShlomoCode: Okay, I got it. I don't think it should be done implicitly, I'll add a separate "cache refresh" function and call it explicitly.

ejbills: Fair enough

ShlomoCode:
> but if I use it while I have an app fullscreen and I have the full size window hover action, the app seems to freak out. Can you reproduce?

No...
https://github.com/user-attachments/assets/fd45fef7-75e1-467d-8748-7fd6da5a9528

ejbills:
> > but if I use it while I have an app fullscreen and I have the full size window hover action, the app seems to freak out. Can you reproduce?
> 
> No...
> 
>  Screen.Recording.2024-08-22.at.7.24.32-trimmed-converted.mov

I mean using DockDoor while inside the fullscreen app

ShlomoCode:
Oh yes it fails to show the fullsize preview but it doesn't work for me with v1.1.5 either

ejbills:
@ShlomoCode what is remaining for this PR? Anywhere I can help out?

ShlomoCode:
[this](https://github.com/ejbills/DockDoor/pull/254#issuecomment-2303628342) not resolved for me yet

ejbills:
@ShlomoCode I think I fixed everything, is it good to go now? Mind helping me by testing it?

DockDoor/Utilities/KeybindHelper.swift
ejbills: I added in the explicit call to updating the methods as discussed, so this discussion is resolved.

ejbills:
@ShlomoCode Also just pushed a really nice fade out effect that makes DockDoor feel really nice to use. Check it out, benefits of the changes I made with the hover detection.

ShlomoCode:
The fade effect is nice, but is:
1. Too long (Maybe make it user-configurable)
2. If you move the mouse diagonally out, there is no fade effect:
https://github.com/user-attachments/assets/66bf9600-85b6-4a90-ae22-54c12c9592d7

DockDoor/Components/MouseTrackingView.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    private func setupTrackingArea() {
+        let options: NSTrackingArea.Options = [.mouseEnteredAndExited, .activeAlways, .inVisibleRect]
+        let trackingArea = NSTrackingArea(rect: bounds, options: options, owner: self, userInfo: nil)
+        addTrackingArea(trackingArea)
+    }
+
+    override func mouseExited(with event: NSEvent) {
+        DispatchQueue.main.async {
+            SharedPreviewWindowCoordinator.shared.hideWindow()
```
ShlomoCode, line 89: You missed the check
```
currentAppUnderMouse == nil || currentAppUnderMouse != lastAppUnderMouse
```
This means that if you stand on an app icon, go to a preview, and return to the icon, the preview will close instead of staying open.

ShlomoCode (commented):
.

ShlomoCode (commented):
--

DockDoor/Components/MouseTrackingView.swift
ejbills: I found that it didn't require this check anymore, when I hover back on the icon it re-appears properly. Adding the check back is trivial, I will iron it out today.

ejbills:
> The fade effect is nice, but is:
> 
> 1. Too long (Maybe make it user-configurable)
> 2. If you move the mouse diagonally out, there is no fade effect:
> 
>  CleanShot.2024-08-31.at.21.04.16.mp4



Not sure what your video is showing in regards to point 2, you can change the delay to the next application in the settings. It appears you are hovering over Xcode, which of course will hide the window for the other application. However, in your video I do see that the opacity of the view does not reset for you when you hover back onto the app. I will look into this.

ShlomoCode:
Yes the video is not good, CleanShot X seems to be having trouble capturing the preview for some reason
What about the length of the effect? He seems too annoying to me right now.

ejbills:
> What about the length of the effect? He seems too annoying to me right now.

I can tie that into a Default easily. 

> CleanShot X seems to be having trouble capturing the preview for some reason

Oh ok, so it is resetting properly?

ShlomoCode:
https://github.com/user-attachments/assets/adad5d11-29dd-4d4f-8e1b-b7c72bc87245


Note that the preview disappears with no effect when you move the mouse over the icon without a preview

ejbills:
> CleanShot.2024-08-31.at.22.11.43.mp4
> 
> Note that the preview disappears with no effect when you move the mouse over the icon without a preview

Fixed

ejbills:
@ShlomoCode try out the new commits, I think it's good to go. I just need to test it on some external monitors. I will merge it later today if everything looks good to you.

ShlomoCode:
> This is intended behavior

This is inconsistent behavior. If the preview disappears with an effect, the effect should always be there

ejbills:
> This is inconsistent behavior. If the preview disappears with an effect, the effect should always be there

I understand the confusion, but let me clarify the two distinct behaviors:

1. Mouse interaction with hover window: When you interact with the hover window and then move the mouse away, there's a fade effect as the window disappears.

2. Changing hovered dock element: When you move quickly from one dock element to another, the hover window hides instantly.

These are intentionally different behaviors designed for different user interactions. The fade effect provides a smooth transition when you're done viewing a window, while the instant hiding allows for quick navigation between dock elements without visual lag.

DockDoor/Views/Settings/MainSettingsView.swift
```diff
@@ -60,7 +60,7 @@ struct MainSettingsView: View {
             Divider()
 
             HStack {
-                Text("Hover Window Open Delay")
+                Text("Preview Window Open Delay")
```
ejbills, line 63: The verbiage "preview" is used for the preview hover actions. Hover window is used throughout the app and codebase to describe what it is, the hover window. Why is this changed?

DockDoor/Views/Hover Window/SharedPreviewWindowCoordinator.swift
```diff
@@ -64,7 +64,7 @@ final class SharedPreviewWindowCoordinator: NSWindow {
         isMovableByWindowBackground = false
         collectionBehavior = [.canJoinAllSpaces]
         backgroundColor = .clear
-        hasShadow = false
+        hasShadow = true
```
ShlomoCode, line 67: This has been changed at https://github.com/ShlomoCode/DockDoor/commit/9c27ffa29f1e88bb57240b89943d0cd07178966e. Is this a deliberate change?

ShlomoCode (commented):
.

ShlomoCode (commented):
.

ShlomoCode (commented):
.

DockDoor/Views/Hover Window/SharedPreviewWindowCoordinator.swift
ejbills: Yes

DockDoor/Views/Settings/MainSettingsView.swift
ShlomoCode: I think for the users "preview" is much clearer than "hover", as it reflects the essence of the window and not its hierarchy, like the window switcher is a "switcher" and not an "overlay"

ShlomoCode:
There are still cases where the preview stays open incorrectly: https://share.cleanshot.com/rRnsHrt99TqdSx2zmCXm

It happens to me from time to time even with normal use.

ejbills:
> There are still cases where the preview stays open incorrectly: https://share.cleanshot.com/rRnsHrt99TqdSx2zmCXm
> 
> 
> 
> It happens to me from time to time even with normal use. I don't think it can be merged yet...

All the issues related to entering the hover preview have been resolved. The issue you are showcasing here is hovering over a dock element, and then leaving the dock area without interacting with the hover preview at all, therefore the mouseExited does not execute. 

At this point, I don't think we can account for this beyond tracking the mouse outside of the dock area. Solving this issue will involve coding another handler which can catch the case the mouse is far away from the dock, and a preview is showing. Thoughts?

ShlomoCode:
> All the issues related to entering the hover preview have been resolved. The issue you are showcasing here is hovering over a dock element, and then leaving the dock area without interacting with the hover preview at all, therefore the mouseExited does not execute.
> 

Good catch! But `handleSelectedDockItemChangedNotification` in `DockObserver.swift` is supposed to handle such a case, so that's probably where the bug lies (maybe it's related to `hoverProcessingTask?.cancel()`). Can you try to check it? I'm afraid I don't have time for that right now.

ShlomoCode:
@ejbills If I set the Fade Out Duration to 0.1 or 0 there is a strange behavior

https://github.com/user-attachments/assets/66a5bdea-8b74-48f3-83f2-4a9daa3ffc72

ejbills:
> `handleSelectedDockItemChangedNotification`



> > All the issues related to entering the hover preview have been resolved. The issue you are showcasing here is hovering over a dock element, and then leaving the dock area without interacting with the hover preview at all, therefore the mouseExited does not execute.
> 
> Good catch! But `handleSelectedDockItemChangedNotification` in `DockObserver.swift` is supposed to handle such a case, so that's probably where the bug lies (maybe it's related to `hoverProcessingTask?.cancel()`). Can you try to check it? I'm afraid I don't have time for that right now.

This bug is present in prod, the issue is exacerbated because the global mouse listener no longer calls the hide window func. The issue isn't the task processing - not exactly sure what the root cause is.

ejbills:
> @ejbills If I set the Fade Out Duration to 0.1 or 0 there is a strange behavior
> CleanShot.2024-09-01.at.01.45.35.mp4

I can reproduce, I'll fix it tomorrow or later.

ejbills:
@ShlomoCode NSTrackingArea seems to have some reliability issues when the mouse quickly crosses its area. My next course of action is another global mouse listener, but it only lives during the lifespan of the preview window (so there wouldn't be constant CPU usage still). But, I'm not sure that's the best solution. 

I have tested everything else, this is the last bug for this PR. Need some help on this issue, I'm not too sure where to go from here.

ejbills:
@ShlomoCode we can get around this by always hiding any window before presenting a new one in the DockObserver, see commit https://github.com/ejbills/DockDoor/pull/254/commits/bbec88ea3c5883ece840dd9a442728f0d972caa9.

This appears to fix the issue, but want your thoughts on it.

ShlomoCode:
> @ShlomoCode It appears moving the mouse quickly makes the `processSelectedDockItemChanged` method execute twice

Maybe we should implement a debounce mechanism?

ShlomoCode:
The TEMPORARY PATCH commit did not solve the previews staying open incorrectly for me. It seems to moderate but not completely resolve it.

ejbills:
> The TEMPORARY PATCH commit did not solve the previews staying open incorrectly for me. It seems to moderate but not completely resolve it.

It is only meant to fix hovering back onto an app, from the apps preview. That was causing it to stay open when the delay was 0.1, or 0. It does not resolve the window staying open when you don't interact with the window at all.

ejbills:
@ShlomoCode I have solved all issues.
- Hovering back onto an application does not hide the window.
- Quickly passing over a view no longer incorrectly displays a lingering window.
- Lingering windows get cleaned up with a temporary mouse polling timer, resolving issues where the mouse does not enter the window preview at all.

I have played around with this, and I think it irons out all the issues we were dealing with. Please give it a test, and let me know what you think. I left some comments in the WindowDismissalContainer to explain this logic.

ShlomoCode:
@ejbills Sounds good! I'll give it a spin and come back

ShlomoCode:
It seems that when there are several open apps adjacent to each other in the Dock, the issue still exists


https://github.com/user-attachments/assets/c91a43b9-14ab-4f5c-9a72-f400f9141dc1

ejbills:
> It seems that when there are several open apps adjacent to each other in the Dock, the issue still exists

This is the case of point 3 on this comment: https://github.com/ejbills/DockDoor/pull/254#issuecomment-2325385556
> Lingering windows get cleaned up with a temporary mouse polling timer, resolving issues where the mouse does not enter the window preview at all.

The dead zone for cleaning up lingering windows is 800 px. If you move your mouse away while a window is lingering, it will hide after the distance is > 800. I actually think this is too large of an area, what do you think is a reasonable area for this? (You can play around with it on line 59 of `WindowDismissalContainer.swift`)

ejbills:
I have tested this all day with no issues, I will be merging. Again, thanks a ton @ShlomoCode, this is a huge and very beneficial PR for DockDoor. Great work.

Issue #211, opened 2024-07-17
Support for macOS Ventura


tech189:
Hi, would love to try this app but sadly I'm stuck on macOS Ventura. There was some mention that it might be possible to bring support to it at some point but I thought it would be good to track it in an issue.

For reference: https://old.reddit.com/r/macapps/comments/1dnfnj8/dockdoor_a_free_and_open_source_dock_peeking/la3zgh0/
> Hi, current version is macos 14 min. I will be looking at lowering that requirement - stay posted.

> Default swift package iirc depends on a swift language minimum which is incompatible with Monterey

illavoluntas:
I think DockDoor uses a new macOS API to capture the previews of windows whose minimum requirement is **macOS 14 (*Sonoma*)**. So for now it wouldn't work in macOS Ventura...

mehapps:
I thought the new screen capture api was ventura? could be wrong

illavoluntas:
Well, I don’t know exactly what API DockDoor uses. Maybe @ejbills could answer that.

ShlomoCode:
We use [`SCScreenshotManager.captureImage`](https://developer.apple.com/documentation/screencapturekit/scscreenshotmanager/4251334-captureimage) which is macOS 14.0 and above, but [here](https://developer.apple.com/videos/play/wwdc2022/10156/) there seems to be an alternative syntax that works in macOS 13

PR #267, opened 2024-08-23
feat: add support for macOS 13 Ventura

ShlomoCode:
## Describe your changes
Added support for macOS 13 Ventura.

closes https://github.com/ejbills/DockDoor/issues/211

## Core Functionality Changes

No change is expected for macOS 14 and above, for older systems support does not include a number of small things such as SmoothGradient ~~or the option to cancel in warning about overlapping elements in the settings~~.

### DIFF
--- a/DockDoor/Components/Marquee.swift
+++ b/DockDoor/Components/Marquee.swift
@@ -31,9 +31,13 @@ struct TheMarquee<C: View>: View {
         if !animating { return }
         let offsetAmount = contentSize.width + spacingBetweenElements
         let duration = offsetAmount / speedPtsPerSec
+
         withAnimation(.linear(duration: duration)) {
             offset = -offsetAmount
-        } completion: {
+        }
+
+        // Simulate the completion handler taht only available on macOS 14.0+
+        doAfter(duration) {
             offset = 0
             doAfter(secsBeforeLooping) {
                 if animating {
@@ -63,8 +67,8 @@ struct TheMarquee<C: View>: View {
         .measure($containerSize)
         .compositingGroup()
         .opacity(measured ? 1 : 0)
-        .onChange(of: containerSize) { _, _ in startAnimation() }
-        .onChange(of: contentSize) { _, _ in startAnimation() }
+        .onChange(of: containerSize) { _ in startAnimation() }
+        .onChange(of: contentSize) { _ in startAnimation() }
         .onAppear { startAnimation() }
         .onDisappear { animating = false }
     }
@@ -76,23 +80,25 @@ extension View {
             if !disable {
                 GeometryReader { geo in
                     DynStack(direction: axis, spacing: 0) {
-                        SmoothLinearGradient(
-                            from: .black.opacity(0),
-                            to: .black.opacity(1),
-                            startPoint: axis == .horizontal ? .leading : .top,
-                            endPoint: axis == .horizontal ? .trailing : .bottom,
-                            curve: .easeInOut
-                        )
-                        .frame(width: axis == .horizontal ? fadeLength : nil, height: axis == .vertical ? fadeLength : nil)
-                        Color.black.frame(maxWidth: .infinity)
-                        SmoothLinearGradient(
-                            from: .black.opacity(0),
-                            to: .black.opacity(1),
-                            startPoint: axis == .horizontal ? .trailing : .bottom,
-                            endPoint: axis == .horizontal ? .leading : .top,
-                            curve: .easeInOut
-                        )
-                        .frame(width: axis == .horizontal ? fadeLength : nil, height: axis == .vertical ? fadeLength : nil)
+                        if #available(macOS 14.0, *) {
+                            SmoothLinearGradient(
+                                from: .black.opacity(0),
+                                to: .black.opacity(1),
+                                startPoint: axis == .horizontal ? .leading : .top,
+                                endPoint: axis == .horizontal ? .trailing : .bottom,
+                                curve: .easeInOut
+                            )
+                            .frame(width: axis == .horizontal ? fadeLength : nil, height: axis == .vertical ? fadeLength : nil)
+                            Color.black.frame(maxWidth: .infinity)
+                            SmoothLinearGradient(
+                                from: .black.opacity(0),
+                                to: .black.opacity(1),
+                                startPoint: axis == .horizontal ? .trailing : .bottom,
+                                endPoint: axis == .horizontal ? .leading : .top,
+                                curve: .easeInOut
+                            )
+                            .frame(width: axis == .horizontal ? fadeLength : nil, height: axis == .vertical ? fadeLength : nil)
+                        }
                     }
                 }
             } else {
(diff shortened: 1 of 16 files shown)


DockDoor/Components/Marquee.swift
```diff
@@ -63,8 +67,8 @@ struct TheMarquee<C: View>: View {
         .measure($containerSize)
         .compositingGroup()
         .opacity(measured ? 1 : 0)
-        .onChange(of: containerSize) { _, _ in startAnimation() }
```
ejbills, line 66: Let's keep these as underscores, the variable is not used.

DockDoor/Utilities/WindowUtil.swift
```diff
@@ -102,9 +124,12 @@ enum WindowUtil {
         config.height = Int(window.frame.height * scaleFactor) / Int(Defaults[.windowPreviewImageScale])
 
         config.showsCursor = false
-        config.captureResolution = .best
```
ejbills, line 105: This is kind of important, why is that removed?

DockDoor/Views/Hover Window/SharedPreviewWindowCoordinator.swift
ejbills, line 1: Does the window switcher work with this? This was the main reason DD is min macOS 14, the observable property allows us to reflect the changes in index when using the switcher. I'm not sure if @Published properties are a slot-in replacement - however I may be wrong.

DockDoor/Utilities/WindowUtil.swift
ShlomoCode: `'captureResolution' is only available in macOS 14.0 or newer`
So we need to condition it to avoid compiler error...
Maybe I'll separate it into a separate "captureWindowModern" function and mark it all as available(macOS 14.0, *), that would be nicer

ejbills:
> option to cancel in warning about overlapping elements in the settings

Is this really not a thing in previous versions? I think we can make the message util take another approach if it's not available, similar to the window manager legacy func you added. 

Cool PR, thank you! Love it!

DockDoor/Components/Marquee.swift
ShlomoCode: I just removed the second parameter as it is not supported in macOS 13, i did not change it to a named parameter

DockDoor/Utilities/WindowUtil.swift
ejbills: Yes, keep in mind the option of specifying a capture resolution has existed for a while. I'm not exactly sure how to set it on the legacy screen API though... but it is possible.

ShlomoCode:
The message still appears, only the "cancel" button will not work (we do need to hide it), but the main goal, to draw the user's attention to the overlap, has been achieved. How can I get the oldValue in macOS 13?

DockDoor/Views/Hover Window/SharedPreviewWindowCoordinator.swift
ShlomoCode: It seems to work fine, although I'm actually not very familiar with the windows switcher. Is there anything special I should look out for?

Uploading CleanShot 2024-08-23 at 04.15.05.mp4…

ejbills:
> How can I get the oldValue in macOS 13?

Store the value in a state var within the settings view.

DockDoor/Views/Hover Window/SharedPreviewWindowCoordinator.swift
ejbills: I'll try it out in a bit

ShlomoCode:
Thanks, done.

ShlomoCode:
@ejbills One thing that is strange to me is that the image appears to be quarter size, can you figure out why this is happening? 
CleanShot 2024-08-23 at 05 03 11@2x

You can restore it in current macOS by changing this:
```swift
let image: CGImage = if #available(macOS 14.0, *) {
    try await captureWindowImageModern(window: window, width: width, height: height)
} else {
    try await captureWindowImageLegacy(for: window, width: width, height: height)
}
```

From 14 to 19 for example...

DockDoor/Utilities/WindowUtil.swift
ShlomoCode: Yes `.bestResolution`, but I already define it

ShlomoCode:
Tested with macOS 13.0 VM, looks fine
I think it's ready

ejbills:
> Tested with macOS 13.0 VM, looks fine I think it's ready

The window switcher does not function properly, pressing tab now does nothing as the @Observable was responsible for making the view react to the index changes.

ejbills:
@ShlomoCode I am working on integrating the observed object, please don't push anything

ejbills:
@ShlomoCode can you give me write permissions for your fork?

ejbills (changes requested):
https://github.com/ejbills/DockDoor/pull/267#issuecomment-2307416825

illavoluntas:
Once this PR is merged, does it mean DockDoor will be fully compatible with Ventura at the next release?

Does the API used for Ventura work on Monterey? Or is it 13.0+?

ShlomoCode:
> Once this PR is merged, does it mean DockDoor will be fully compatible with Ventura at the next release?
> 

Yes

> Does the API used for Ventura work on Monterey? Or is it 13.0+?

I can adapted quite easily to macOS 12.3, the question is how much demand is there for it?

ShlomoCode:
@ejbills Sorry for the late reply, you should now be able to push. Pay attention to do a rebase pull first  (`git pull --rebase`)

illavoluntas:
> I can adapted quite easily to macOS 12.3, the question is how much demand is there for it?

There will always be demand for older system compatibility. As of today, I think some people still use 10.15 Catalina, it might be the oldest most demanded version (macOS 11 Big Sur abandoned a lot of Macs).

ShlomoCode:
OK, let's merge Ventura first.

ShlomoCode:
@ejbills I see that fixing the broken onChange currIndex is a small change, so I just pushed it

ejbills:
> @ejbills Sorry for the late reply, you should now be able to push. Pay attention to do a rebase pull first (`git pull --rebase`)



> @ejbills Sorry for the late reply, you should now be able to push. Pay attention to do a rebase pull first (`git pull --rebase`)

My local changes actually get rid of the coordinator being a singleton. I rewrote the logic to ensure it properly makes use of the ObservableObject pattern for passing data to child views. So, the onChange won't be necessary. I will push later today, thanks!

ShlomoCode:
It sounds more appropriate for a separate refactor commit

ejbills:
> It sounds more appropriate for a separate refactor commit

 I don't think a separate refactor commit is necessary in this case. The change to an ObservableObject pattern is directly related to the PR's purpose of adding macOS Ventura support. It's a relatively small change, impacting less than ~100 lines of code. I'll be pushing these changes soon as part of this PR, as I am away from my computer at the moment. 

Further refactoring is on my todo list, and it will not be in this PR. Hope that addresses your concerns.

ShlomoCode:
@ejbills You dropped some of my code...
CleanShot 2024-08-25 at 03 49 37@2x
CleanShot 2024-08-25 at 03 49 29@2x

When working on PR together, it is advisable to wait for a mutual review

PR #321, opened 2024-09-14
chore: update rendering of site to new layout

ejbills:
image


ejbills:
@illavoluntas thoughts? Doesn't work on mobile.

illavoluntas:
> @illavoluntas thoughts? Doesn't work on mobile.

Why not? Please give me a few hours/a couple days so that I can see in detail what you changed. 
I have a suggestion: shouldn’t the logo appear one with a greater width?

illavoluntas:
> @illavoluntas thoughts? Doesn't work on mobile.

Well, what technology did you use? Did you write this code by yourself?

I have a few remarks:
- Some paragraphs are too small (especially on the FAQ).
- The new layout looks fine, but is unstable... That's why I chose buttons which you have to click in order to make it work on mobile.
- Some fonts are too thin, I don't think it's good for accessibility.
- The download section disappeared, we can't only rely on a dropdown menu which opens on hover (again, motor disorder accessibility, screen readers)
- Logo is too small.
- Although incorporating text inside the HTML saves from JavaScript fails, you will have to update it in addition to the JSON everytime a key changes, instead of just modify the `locales/en/translation.json` file.
- Everytime I tried sticky headers, it failed and resulted in buggy situations... This one seems to work, but yeah, don't forget mobile phones.
- For mobile phones, you would have to integrate a hamburger menu for the top and make sure items are big enough.
- If you zoom in a little, you can see that `Download` and `Language` do not use system font; they use the default Arial instead.
image

PR #406, opened 2024-12-22
fix: window dragging gif not adapted in index.html

illavoluntas:
Added class "responsive-img" like all other images so that it fits in the screen


ejbills (approved):

ejbills:
Thank you

PR #546, opened 2025-05-11
feat: redesigned settings

### LINKED ISSUE #446: [FR] Give user separate Window Preview Layout Limit settings (row/col) for Dock Previews vs Window Switcher functionality
## Problem or Need
Current single setting makes it impossible to have a single row of dock peviews and a grid of window switcher previews (for example, 5x5).  I prefer a single row for dock preview but a grid for window switcher.  I can't find any combination of row/col values that allow for this.

From my testing it appears that the row value takes precedence over the col value, so if I set it to 1 row and 5 columns but I have 8 documents open in Numbers for instance, it will show more than 5 columns but clamp to a single row.  This is fine but may not be clear to all users.

## Proposed Solution
Separate settings for Window Preview Layout Limits (row/col) for Dock Preview and Window Switcher functionality.

I would recommend a new Dock Preview tab that holds settings specific to Dock Preview functionality, including some settings from the General tab such as 'Prevent dock from hiding during preview' and 'Keep previews visible during lateral movement'.  Window Switcher tab could have its own Layout Limits settings as this tab is fairly sparse right now

You could even go so far as allowing the user to turn off Dock Previews and just use Window Switcher functionality via a master checkbox on a new Dock Preview tab, like Window Switcher has today (but this is not directly relevant to this FR).

## Alternative Solutions
None

## Additional context
None

## Potential Impact
Slightly more complex settings but more user delight

## macOS Version
15.2

## App Version
1.6

- [x ] I have reviewed the open issues to ensure this is not a duplicate feature request

### LINKED ISSUE #448: [FR] Default settings could match Windows OS
I’m a big fan of this app and believe it’s a must-have for every Mac user.

After installing it, I usually make a few tweaks to make it behave more like the default Windows OS. I suspect this is what most Mac users who install this app are looking for.

Specifically I believe this roughly matches the default Windows behaviour:

1. Prevent dock from hiding during previews
2. Disable "Sort Window Previews by Date"
3. Dock Preview Window Hover Action: Present a full size preview
4. Preview Hover Delay = 0.2 seconds seems to match Windows the best?
5. Window Preview Size = 1/6 seems to match Windows the best?

Considering this, I think it would be an interesting option to install the app with as close to Windows OS defaults as possible.

Alternatively, you could provide a choice on the first startup and in the settings, allowing users to select between “Use Windows presets” or “Use DockDoor defaults.”

I’m suggesting this because I’ve been recommending this app to all my Mac user friends, and they all want it to work like Windows (since that’s what they’re used to otherwise). Currently, this app requires a bit of setup to achieve that and some 'non-techy' users may not want to delve into the settings.

Thanks!

apotenza92:
## Describe your changes
New settings menu as per discussion in [448](https://github.com/ejbills/DockDoor/issues/448).
closes #446 closes #448

#448 
#446

- [] If this change affects core functionality, I have added a description highlighting the changes
- [] I have followed conventional commit guidelines

## Core Functionality Changes
- Redesigned settings screen is a major change


ejbills:
@apotenza92 looks like there are some merge conflicts. If you need my help to resolve this, let me know. I am going to commit some changes to this PR anyways, I like the idea of the images on the top of the settings. I want to make them live previews. 

There are also some other settings I want to include to the main settings which I think are required.

DockDoor/consts.swift
```diff
@@ -73,23 +73,6 @@ extension Defaults.Keys {
 
 // MARK: Display Configurations
 
-enum WindowTitleDisplayCondition: String, CaseIterable, Defaults.Serializable {
```
ejbills, line 76: @apotenza92 let's restore this functionality

DockDoor/Views/Settings/DockPreviewsView.swift
```diff
@@ -0,0 +1,227 @@
+import Defaults
+import SwiftUI
+
+// Add enum for preview size
+private enum DockPreviewSize: String, CaseIterable, Identifiable {
```
ejbills, line 5: Move this to consts.swift where the other declarations are of settings configurations

ejbills:
Also, let's restore the keyboard shortcuts controls

DockDoor/consts.swift
apotenza92: No worries, I think I wanted it to become an independent setting for the previews and switcher but I may have implemented it wrong or deleted this by accident.

ejbills:
CleanShot 2025-05-16 at 11 38 14@2x
Thoughts?

ejbills:
Closing in favor of https://github.com/ejbills/DockDoor/pull/549, this PR was a large source of inspiration, however there was a large amount of work I wanted to do and it was just easier to make a new PR. I will credit you in the update notes.

Thanks!!

PR #549, opened 2025-05-16
feat: redesign settings

### LINKED ISSUE #283: chore: Fix several strings
Strings:
1. `• Allows for enhanced visual information in the dock`
2. `• Enables real-time interaction with dock items`
3. `• To detect when you hover over the dock`
4. `Adjust this if the preview is misaligned with dock`
5. `Enhance your dock experience!`
6. `If you need to access the menu bar icon, launch the app to reveal it for 10 seconds.`
7. `Press any key combination after holding the initialization key to set the keybind.`
8. `Show Window Title in Previews`
9. `Show Window Titles on Previews`
10. `Use default MacOS keybind ⌘ + ⇥`

need to be modified in my opinion.

- Iterations of `dock` need to be updated to `Dock` (1, 2, 3, 4, 5).
- String 6 needs to be updated for new method. Suggested new string: `To access this window again, simply launch the app and it will open automatically.`
- String 7 is unclear: user only needs to press *one* key, not holding the initialization key at the same time.
- Strings 8 and 9 are quite identical.
- `MacOS` needs to be updated to `macOS` (10).

Furthermore, in other strings, what is the meaning of putting a capital letter at the beginning of each word? Looks quite strange to me.

### LINKED ISSUE #446: [FR] Give user separate Window Preview Layout Limit settings (row/col) for Dock Previews vs Window Switcher functionality

### LINKED ISSUE #448: [FR] Default settings could match Windows OS

### LINKED ISSUE #477: [FR] [FR] Hiding application name on window preview
## Problem or Need
In the window preview, when an application has a small window size the name of the application isn't fully shown.
Image

## Proposed Solution
Can you add the option of hiding the application name so that only the icon of the application is visible.

## macOS Version
macOS Sequoia 15.2 (24C101)

## App Version
1.6.1

- [x] I have reviewed the open issues to ensure this is not a duplicate feature request

ejbills:
## Describe your changes
New settings menu as per discussion in #448
closes #446 closes #448 closes #477 closes #283

#448 
#446

- [x] Update documentation
- [x] Verify window titles are updated
- [x] Ensure no regression with WindowProvider change to window info objs
- [x] Drag testing

CleanShot 2025-05-16 at 16 29 54@2x
CleanShot 2025-05-16 at 16 30 00@2x
CleanShot 2025-05-16 at 16 30 05@2x
CleanShot 2025-05-16 at 16 30 09@2x


ejbills:
would like some feedback @ShlomoCode @apotenza92

DockDoor/Views/Settings/Help/BaseSettingsView.swift
```diff
@@ -0,0 +1,20 @@
...
+    let content: Content
+
+    init(@ViewBuilder content: () -> Content) {
+        self.content = content()
+    }
+
+    var body: some View {
+        ScrollView {
+            content
+                .padding(20) // Apply padding to the content itself inside the ScrollView
```
ejbills, line 13: remove comments

DockDoor/Views/Settings/WindowSwitcherSettingsView.swift
```diff
@@ -1,5 +1,5 @@
 import AppKit
-import Carbon
+import Carbon // For key codes, if still used by KeyCodeConverter, though UserKeyBind uses UInt16
```
ejbills, line 2: remove comments

DockDoor/Views/Settings/WindowSwitcherSettingsView.swift
```diff
@@ -32,264 +31,38 @@ struct KeyCapView: View {
...
+
     override func viewDidAppear() {
         super.viewDidAppear()
-        view.window?.makeFirstResponder(view)
+        if coordinator?.parent.isRecording ?? false {
+            DispatchQueue.main.async { // Ensure UI updates on main thread
+                self.view.window?.makeFirstResponder(self.view)
+            }
+        }
     }
```
ejbills, line 53: test key bond

DockDoor/Views/Settings/WindowSwitcherSettingsView.swift
```diff
@@ -32,264 +31,38 @@ struct KeyCapView: View {
...
 
     override func keyDown(with event: NSEvent) {
         coordinator?.handleKeyEvent(event)
     }
 
     override func flagsChanged(with event: NSEvent) {
-        coordinator?.handleKeyEvent(event)
+        // Modifiers are handled by the Picker in MainSettingsView
+        // coordinator?.handleKeyEvent(event)
```
ejbills, line 60: delete

DockDoor/Utilities/Window Management/WindowUtil.swift
```diff
@@ -491,22 +494,23 @@ enum WindowUtil {
...
+                        var mutableCopy = updatedWindow
+                        mutableCopy.image = image
+                        updateDesktopSpaceWindowCache(with: mutableCopy)
                     }
                 }.value
             }
             return
         }
 
-        guard window.owningApplication != nil
```
ejbills, line 508: restore 99 % of this file which is just removing comments? keep code changes

DockDoor/Views/Hover Window/WindowPreviewHoverContainer.swift
```diff
@@ -484,25 +457,20 @@ struct WindowPreviewHoverContainer: View {
...
     private func checkForShakeGesture(currentPoint: CGPoint) -> Bool {
         let now = Date()
-        // Check at most every 50ms to maintain responsiveness
         guard now.timeIntervalSince(lastShakeCheck) > 0.05 else { return false }
         lastShakeCheck = now
 
-        // Add new point to tracking array
         dragPoints.append(currentPoint)
 
-        // Keep only last 20 points to analyze recent movement
```
ejbills, line 494: restore

DockDoor/Views/Settings/BaseSettingsView.swift
```diff
@@ -0,0 +1,20 @@
...
+        self.content = content()
+    }
+
+    var body: some View {
+        ScrollView {
+            content
+                .padding(20) // Apply padding to the content itself inside the ScrollView
+        }
+        // This frame makes the content area of each tab uniform.
+        // It allows the width to expand if the window is wider, but fixes the max height.
```
ejbills, line 16: remove comments

oliverphardman:
Hello 👋 I am broadly down with these changes, but a lot of assumptions are being made here. There is no telemetry data showing how many users do bother to change settings and what they change them to.

Over the long-term, I think it's worth anonymously collecting this stuff. I appreciate that's probably a cost for someone to cover. I'd hazard a guess that most users aren't developers and aren't going to take the time to raise an issue.

ejbills:
> Hello 👋 I am broadly down with these changes, but a lot of assumptions are being made here. There is no telemetry data showing how many users bother to change settings and what they change them to.
> Over the long-term, I think it's worth anonymously collecting this stuff. I appreciate that's probably a cost for someone to cover. I'd hazard a guess that most users aren't developers and aren't going to take the time to raise an issue about something they don't like.

I think that is a discussion for a future addition, however for the scope of this PR I feel that we cannot act on users that do not engage and/or provide feedback. The changes are largely based on the discussion with active users in https://github.com/ejbills/DockDoor/issues/448.

In regards to telemetry, it would indeed be useful for development, but then I would be the only one who could access them. I would go down that path if DockDoor was profit-driven, however I simply do not have the time to manage the infrastructure and additional code changes that it would require, at least in relation to this PR. I would be interested to see how other FOSS apps engage with anonymized telemetry data. This feature would have to be opt-in only however, as I am against tracking in general.

PR #719, opened 2025-08-30
Respect inactivity timer when hovering empty apps with lateral movement off

ezamosch:
Added behavior:

- When “Keep previews…” is turned off, hovering a Dock icon for an app with no active windows no longer triggers an immediate hideWindow(). Instead, DockObserver sets the requireInactivityDismissal flag and hands off to WindowDismissalContainer, where the “Preview Window Inactivity Timer” fires and the preview fades out.

- After the fade completes, WindowDismissalContainer calls hideWindow(), and then DockObserver.resetTrackingAfterContainerDismissal(), preserving the required reset of tracking (lastAppUnderMouse, etc.).

- If the user re-enters the preview area, the fade is correctly canceled (as before).

### DIFF
--- a/DockDoor/Utilities/DockObserver.swift
+++ b/DockDoor/Utilities/DockObserver.swift
@@ -50,6 +50,11 @@ final class DockObserver {
 
     private var eventTap: CFMachPort?
 
+    // When true, the WindowDismissalContainer should treat the dock icon as "not hovered"
+    // to allow the inactivity timer fade-out to dismiss the preview window.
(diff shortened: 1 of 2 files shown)


ezamosch:
I did try to remake as you said, but couldn't. It only worked for me with extra flags, otherwise something always go wrong, I mean something stops work properly.

ejbills:
I have implemented this in the latest beta `1.25.3.5_pre_1_25_4`. Please try it out, and let me know what you think.

PR #777, opened 2025-10-04
feat: WindowPreview fixed dimensions and oldest first

bcov77:
## Describe your changes

I've got two toggleable changes I'm porting from "another app" to this one:

- Fixed-dimensions for the previews: In horizontal mode, the previews extend out to max columns before starting a new row (and visa versa in !horizontal mode)
- Show oldest windows first: This mostly applies to when you aren't sorting by date. The top-left window will always be the oldest window that you opened

The "reason" why these two changes are needed for me is that I use the preview pane to remember which window is which (I'm not much of a tabs person). I'll frequently have 14 terminals open and 20 sublime windows and the only way I can keep track of them is that they are in a persistent order within the preview pane.

See below I think I also fixed a bug in image caching.

## Core Functionality Changes

My two options affect the dimensioning of the preview pane as well as the ordering of the windows. But both of the changes are in toggles that are disabled by default.

--------

Lastly, I think I found and fixed a bug with how the image caching works. From what I can tell, `WindowInfo.lastAccessedTime` is only updated whenever a window is brought to the front (through calls to `WindowInfo.updateWindowDateTime()` and `WindowInfo.updateTimestampOptimistically()`). However, `lastAccessedTime` is also used to determine if the image cache is out of date. Updating the image (which is an operation that's triggered more frequently that it being brought to the front) doesn't seem to update `lastAccessedTime` (`updateDesktopSpaceWindowCache()` discards it if the window already existed) and so in principle, most of the cached images are overwritten every time they are requested.

I fixed this bug(?) by splitting`WindowInfo.lastAccessedTime` out into `WindowInfo.lastAccessedTime` and `WindowInfo.lastImageCaptureTime`. I do know that if `!sortWindowsByDate` the cache time never updated.

### DIFF
--- a/DockDoor/Views/Settings/MainSettingsView.swift
+++ b/DockDoor/Views/Settings/MainSettingsView.swift
@@ -836,6 +836,7 @@ struct MainSettingsView: View {
                 let qualityDefault = PreviewQualityProfile.standard.settings
                 screenCaptureCacheLifespan = qualityDefault.screenCaptureCacheLifespan; windowPreviewImageScale = qualityDefault.windowPreviewImageScale
                 bufferFromDock = Defaults.Keys.bufferFromDock.defaultValue; sortWindowsByDate = Defaults.Keys.sortWindowsByDate.defaultValue; shouldHideOnDockItemClick = Defaults.Keys.shouldHideOnDockItemClick.defaultValue; dockClickAction = Defaults.Keys.dockClickAction.defaultValue; previewHoverAction = Defaults.Keys.previewHoverAction.defaultValue; aeroShakeAction = Defaults.Keys.aeroShakeAction.defaultValue
+                showOldestWindowsFirst = Defaults.Keys.showOldestWindowsFirst.defaultValue
 
                 showMenuBarIcon = Defaults.Keys.showMenuBarIcon.defaultValue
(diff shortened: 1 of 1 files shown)


bcov77:
Screenshot showing `previewFixedDimensions` doing its thing: (with `showOldestWindowsFirst` also active)
image

Screenshot showing `previewFixedDimensions`'s checkbox:
image

Screenshot showing `showOldestWindowsFirst`'s checkbox:
image

DockDoor/Utilities/Window Management/WindowUtil.swift
```diff
@@ -714,9 +722,10 @@ enum WindowUtil {
             {
                 let updatedWindow = existingWindow
                 await Task.detached(priority: .userInitiated) {
-                    if let image = try? await captureWindowImage(window: actualSCWindow, forceRefresh: true) {
+                    if let (image, imageDate) = try? await captureWindowImage(window: actualSCWindow, forceRefresh: true) {
```
bcov77, line 725: I'm new to swift. Does this if-statement still work correctly?

ejbills: Yes this works

ejbills:
Closing this as inactive - this pull request has drifted significantly from the code currently on `main`. Please feel free to re-open your PR when you have time to address the feedback and merge conflicts.

PR #837, opened 2025-11-15
feat: Add CMD + Right Click to quit apps from dock

danieletieghi:
Added new feature to allow users to quickly quit applications directly from the dock using CMD + Right Click on the app icon. This provides a faster alternative to opening the context menu and selecting quit (this was a feature that I used to love from HyperDock)

Features:
- CMD + Right Click on dock icon terminates the application
- CMD + Option + Right Click performs force quit
- Event is consumed to prevent context menu from appearing

Implementation:
- Modified event tap to capture both left and right mouse clicks
- Added handleCmdRightClickQuit() method to handle quit action
- Automatically purges app cache after termination

## Core Functionality Changes

### DIFF
--- a/DockDoor/Utilities/DockObserver.swift
+++ b/DockDoor/Utilities/DockObserver.swift
@@ -549,9 +549,7 @@ final class DockObserver {
 
     private func handleCmdRightClickQuit(app: NSRunningApplication, event: CGEvent) {
         Task { @MainActor in
-            let shouldForceQuit = event.flags.contains(.maskAlternate)
-
-            if shouldForceQuit {
+            if event.flags.contains(.maskAlternate) {
                 app.forceTerminate()
             } else {
                 app.terminate()


DockDoor/Utilities/DockObserver.swift
```diff
@@ -539,4 +549,20 @@ final class DockObserver {
...
+            // Check if Option key is also pressed for force quit
+            let shouldForceQuit = event.flags.contains(.maskAlternate)
+
+            if shouldForceQuit {
+                app.forceTerminate()
+            } else {
+                app.terminate()
+            }
+
+            // Purge the app cache
```
ejbills: This is not required as it will be called by the termination observer - thanks for the PR!

DockDoor/Utilities/DockObserver.swift
```diff
@@ -539,4 +549,20 @@ final class DockObserver {
             }
         }
     }
+
+    private func handleCmdRightClickQuit(app: NSRunningApplication, event: CGEvent) {
+        Task { @MainActor in
+            // Check if Option key is also pressed for force quit
+            let shouldForceQuit = event.flags.contains(.maskAlternate)
+
+            if shouldForceQuit {
```
ejbills: Since shouldForceQuit is initialized but not used anywhere else, can we move the `event.flags.contains(.maskAlternate)` call to be inline with the `if` statement?

PR #861, opened 2025-12-01
feat: added support for showing a customizable indicator line below the currently active app

maddada:
Hi there,

Thank you for making this insanely awesome app and making it free and open source! I'll be sure to donate and keep supporting it.

I just recently started using it instead of sidebar due to some issues I had with that app but I noticed that there isn't a way to show an indicator for the active app.

This PR just adds the first basic part which is an indicator below the app icon.

2025-12-01_CleanShot_05-08-21

Also adds an indicator border on the active window in the preview window.

2025-12-01_Cursor_05-44-08

Please feel free to adjust it as you see fit. There's 2 commits, first one adds the indicator.
Second one adds basic support for auto hiding docks by adding a fade out animation and some sliders for testing different animation values and hover distance to activate. 

Up to you if you want to hide or keep these sliders visible to the user. Also up to you to set the actual values for these as I never use auto-hide dock.

2025-12-01_Vivaldi Snapshot_05-17-13

Small issues I noticed:
1- When apps are closed or opened the over active app indicator isn't shifting so it can get slightly unaligned in some cases.
2- If the user moves the dock to the left or right then the last active app indicator will remain visible until they switch the active app.

Both of the ones above can be solved by refreshing the indicator every x seconds. Please tell me if this is what you want to do, or if we should just ignore it since they're minor issues and adding polling isn't great.

3- It's not possible to use this for dock on left/right sides yet. Personally don't use those, but please tell me if it's a must to implement them to get this PR in and I'll work on those.

4- I'm using a naive approach for detecting if the dock is shown (if auto hide is enabled), so if the user goes to hover on the preview window then the active app indicator will go away.

Can you please guide me on how to make this more robust or implement a more robust solution for this part so we don't rely just on px height?


ejbills:
Thanks for the awesome PR! This is a great idea, and I think with some collaboration we can make the feature complete for left and right docks, since I use the dock on the left. 

Some notes about requiring polling, which I think we can avoid entirely:
> 1- When apps are closed or opened the over active app indicator isn't shifting so it can get slightly unaligned in some cases.

We can notify the ActiveAppIndicatorCoordinator of a shift in the dock as DockDoor already listens and responds to applications being opened or quit. This is part of the code base here: https://github.com/ejbills/DockDoor/blob/74ee2eb7e557a02f11bd8bf36d978ee1c1b9b148/DockDoor/Utilities/Window%20Management/WindowManipulationObservers.swift#L69-L90

There are also a couple of other instances of the dock shifting in size, like windows being minimized if the user has this option disabled: 
CleanShot 2025-11-30 at 18 07 57@2x
With this, the window being minimized is added to the right side of the dock. Again, you can notify the singleton of this action in the same observer code I posted above.
> 2- If the user moves the dock to the left or right then the last active app indicator will remain visible until they switch the active tab.

The dock positioning can be wonky to work with, but it is possible to determine the orientation of it without CoreDock (the private Dock API DockDoor uses): https://github.com/nate-parrott/ball/blob/main/Ball/DockUtils.swift#L11-L21
I *think* that you could theoretically observe Dock orientation changes by listening to `NSApplication.didChangeScreenParametersNotification`. When the Dock position changes, this notification is fired because the main screen's visibleFrame (which excludes the space occupied by the Dock) depends on the Dock's position. With this notification, it would be trivial and inexpensive to validate the orientation changed by tracking the original dock orientation state on app launch, and then comparing it to the latest value computed on the `didChangeScreenParametersNotification` event by calling `DockUtils.getDockPosition()`.

> 3- It's not possible to use this for dock on left/right sides yet. Personally don't use those, but please tell me if it's a must to implement them to get this PR in and I'll work on those.

What makes this the case?

>4- I'm using a naive approach for detecting if the dock is shown (if auto hide is enabled), so if the user goes to hover on the preview window then the active app indicator will go away.
>Can you please guide me on how to make this more robust or implement a more robust solution for this part so we don't rely just on px height?

We can entirely avoid polling and whatnot related to `setupDockVisibilityTracking`.  We can determine WHEN to show by integrating this with the `DockObserver` which fires when an app icon is hovered in the dock. This is how DD determines to render a window preview, and where to render it based on the icon rect. 

You could determine when a user has 'left' the dock (i.e., when dock hiding is enabled, and the mouse has left the dock icon). This fires an additional `processSelectedDockItemChanged` (as the dock item has changed from having a value to no longer being hovered) which we can then validate the user has left like so: https://github.com/ejbills/DockDoor/blob/74ee2eb7e557a02f11bd8bf36d978ee1c1b9b148/DockDoor/Views/Hover%20Window/Shared%20Components/SharedPreviewWindowCoordinator.swift#L656-L674

Please let me know if you have any questions about these ideas. Currently, constant mouse polling is computationally expensive and isn't worth it for the active indicator. But, it is definitely possible to implement this without polling.

maddada:
Hey there. So I worked on the issues we spoke about.

I've added support for left and right positions for the dock.

2025-12-01_CleanShot_17-25-51@2x

I also improved the the implementation to not require polling at all anymore. It relies only on listeners now.

Finally, the position of the indicator is staying in sync when apps minimize, open or close, or when dock position is changed. 

However, I still have one issue remaining which is when you stop hovering on the dock but don't pass on the preview first, then the indicator remains visible. I'm not sure what the best solution is for this issue. Please check it out if you can.



Another note is that the user needs to enable ″Prevent dock from hiding during previews" in to avoid the app indicator being visible while the dock hides which looks strange. Maybe it should be auto enabled + toggle disabled if user enables the active app indicator?

2025-12-01_DockDoor_15-56-53

maddada:
Thinking more about it, since the gist you shared gives the size of the dock, I'll try to set the height, width, offset of the indicator automatically to remove those unnecessary sliders.

maddada:
Added auto toggle for the offset and height (recommended to be enabled, no issues with this one)

Also added auto toggle for the width of the indicator. It's not centered sometimes though depending on the width of the icons, so I added a "Shift Indicator" slider to align in.

Feel free to modify the settings as you wish.

Only issue remaining is the one with auto hide that I mentioned above, other than that it's working pretty well IMO. 

Please let me know if there's anything else.

2025-12-04_DockDoor_03-31-10

ejbills:
Will review some more this weekend, thanks for all your effort here!

ejbills:
Can I have write access to your fork? I will fix the merge conflicts when I review and merge it tomorrow.

maddada:
@ejbills I invited you to the repo, you should have write access after you accept.

Fixed the merge conflict since it was very small.

ejbills:
This works great with auto-hide disabled. Would you be okay if we document that auto-hide isn’t supported for this feature? The workarounds are tricky to get right, and I’d rather ship something solid for the common case.

ejbills:
Also, I moved the new active window border color to an appearance settings option, default to off.

maddada:
Yes, sure sounds good to say unsupported for auto hiding dock for now.

Thank you for simplifying the code. I'm sorry it wasn't the best but I'll try to learn from your implementation.

PR #876, opened 2025-12-05
feat: add live window preview support using ScreenCaptureKit

suatkocar:
## Description
Add real-time video streaming for window previews using ScreenCaptureKit instead of static screenshots.

## Features
- Live video preview using ScreenCaptureKit streams
- Configurable quality: Standard, High, Retina (2x)
- Configurable frame rate: 30 FPS, 60 FPS, 120 FPS (ProMotion)
- Settings toggle in General > Preview Appearance & Quality
- Works in both dock previews and full-size preview mode

## Changes
- **New file**: `LiveWindowCapture.swift` - Manages SCStream instances per window
- **consts.swift**: Added `LivePreviewQuality` and `LivePreviewFrameRate` enums with settings keys
- **MainSettingsView.swift**: Added UI controls for live preview settings
- **WindowPreview.swift**: Use live captured images when available
- **WindowPreviewHoverContainer.swift**: Start/stop capture on lifecycle events
- **FullSizePreviewView.swift**: Use live captured images for full-size preview

## Testing
1. Enable "Live Preview (Video)" in Settings > General > Preview Appearance & Quality
2. Hover over dock icons to see live window content
3. Test different quality/FPS settings
4. Test full-size preview hover action

### DIFF
--- a/DockDoor/Utilities/Window Management/LiveWindowCapture.swift
+++ b/DockDoor/Utilities/Window Management/LiveWindowCapture.swift
@@ -1,45 +1,59 @@
 import Cocoa
 import Defaults
 import ScreenCaptureKit
+import SwiftUI
 
-@MainActor
-class LiveWindowCapture: NSObject, ObservableObject {
-    static let shared = LiveWindowCapture()
+struct LivePreviewImage: View {
+    let windowID: CGWindowID
+    let fallbackImage: CGImage?
 
-    @Published var capturedImages: [CGWindowID: CGImage] = [:]
+    @StateObject private var capture: WindowLiveCapture
 
-    private var streams: [CGWindowID: SCStream] = [:]
-    private var streamOutputs: [CGWindowID: StreamOutput] = [:]
-    private var isCapturing = false
-
-    var isEnabled: Bool {
-        Defaults[.enableLivePreview]
+    init(windowID: CGWindowID, fallbackImage: CGImage?) {
+        self.windowID = windowID
+        self.fallbackImage = fallbackImage
+        _capture = StateObject(wrappedValue: WindowLiveCapture(windowID: windowID))
     }
 
-    override private init() {
+    var body: some View {
+        Group {
+            if let image = capture.capturedImage ?? fallbackImage {
+                Image(decorative: image, scale: 1.0)
+                    .resizable()
+            }
+        }
+        .task {
+            await capture.startCapture()
+        }
+    }
+}
+
+@MainActor
+class WindowLiveCapture: NSObject, ObservableObject {
+    @Published var capturedImage: CGImage?
+
+    private let windowID: CGWindowID
+    private var stream: SCStream?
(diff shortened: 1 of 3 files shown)


suatkocar:
Hi @ejbills,

Thank you for creating DockDoor - it is an incredibly polished and well-architected app. The codebase is clean and easy to follow, which made contributing straightforward.

## About this PR

This adds live window preview support using ScreenCaptureKit. Instead of static screenshots, users can now see real-time video of their windows.

### Features

- Uses SCStream for efficient per-window capture
- Configurable quality: Standard, High, Retina (2x scale)  
- Configurable frame rate: 30/60/120 FPS (ProMotion support)
- All settings are toggleable in General > Preview Appearance & Quality

### Where it works

- Dock icon hover previews
- Window Switcher (Alt/Cmd+Tab keyboard shortcut)
- Full-size preview on hover action

Since WindowPreview and WindowPreviewHoverContainer are shared components, live preview automatically works everywhere previews are shown.

### Implementation notes

- LiveWindowCapture is a singleton that manages stream lifecycle
- Streams start on onAppear and stop on onDisappear to save resources
- Falls back to static images gracefully when live preview is disabled

I tried to follow the existing code patterns and conventions. Let me know if you would like any changes or have questions about the implementation.

Happy to collaborate on any refinements.

suatkocar:
Hi @ejbills, PR is ready for review. I've also synced with the latest main and resolved the merge conflict.

DockDoor/Views/Hover Window/WindowPreviewHoverContainer.swift
```diff
@@ -169,6 +171,15 @@ struct WindowPreviewHoverContainer: View {
...
         .onAppear {
             loadAppIcon()
+            startLiveCapture()
+        }
+        .onDisappear {
+            Task {
+                await liveCapture.stopAllCaptures()
+            }
+        }
+        .onChange(of: previewStateCoordinator.windows) { _ in
```
ejbills: Is this call required? I feel like `onAppear` would cover every use case of this feature. 

Maybe it would be a better approach to make a more scoped version of `startLiveCapture` to take a single window id, and then we can attach the `onAppear` to the child WindowPreview's directly. This would cover if say, the coordinator had windows appended to it while the hover container is already visible. The child WindowPreview would call its own `onAppear` and would start the stream.

ejbills:
I noticed that if a window is minimized or hidden prior to DD launching, the live preview does not fall back and therefore fails to show any image. 
CleanShot 2025-12-07 at 15 41 00@2x

ejbills:
I refactored this to move away from the singleton pattern. The main issue was that we had to manually coordinate start/stop calls and track everything in a shared dictionary, which felt clunky. The new approach has each WindowPreview create its own LivePreviewImage view. This also means no resources are used when the feature is disabled, and each window only observes its own capture state instead of a shared dictionary.

PR #890, opened 2025-12-08
feat: improve window switcher performance

suatkocar:
## Description

Improve Window Switcher responsiveness and add mouse hover selection support. This PR focuses on reducing perceived latency and providing a more fluid experience for users who prefer instant feedback.

## Performance Improvements

- **Add Shift key hold cycling** - Tab hold cycling existed but Shift did not; now both work consistently
- **Implement keyboard repeat timers** - Proper initial delay (0.4s) and repeat interval (0.05s) for smooth cycling
- **Remove 0.175s animation delay** - Eliminated unnecessary delay in hover state changes
- **Disable animations on index changes** - Instant response when cycling through windows
- **Reduce bringToFront retry sleep** - 50ms → 20ms for faster window activation
- **Add thread safety** - Main thread check for AX operations

## Mouse Hover Selection

- **Dual index system** - Separate `focusedIndex` (keyboard) and `hoveredIndex` (mouse) tracking
- **Movement threshold detection** - Prevents accidental hover selection on switcher open
- **Visual differentiation** - Keyboard selection (full opacity) vs mouse hover (50% opacity)
- **Activity-based precedence** - Last input type (keyboard/mouse) determines active selection
- **Timer coordination** - Stops keyboard repeat timers when mouse takes over

## Embedded Window Switcher Controls

- Add option to embed traffic light buttons inside preview frames (Settings toggle)
- Buttons appear on mouse hover for a cleaner, less cluttered look

## Traffic Light Hover Effects

- Per-button hover state tracking
- Dim overlay effect (25% opacity) on hover for better visual feedback

## Tab Filtering

- Add "Show standard tabs as windows" toggle
- Filter out tabbed windows from Window Switcher when disabled

## UX Improvements

- **Filters tab loading spinner** - Async app scanning prevents UI freeze with many installed apps
- Add descriptive text for new settings options

## New Settings

| Key | Default | Description |
|-----|---------|-------------|
| `enableMouseHoverInSwitcher` | `true` | Enable mouse hover selection |
| `useEmbeddedWindowSwitcherElements` | `false` | Embed controls in preview frames |
| `showTabsAsWindows` | `false` | Show browser tabs as separate windows |

## Testing

- [x] Window Switcher keyboard cycling (Tab/Shift+Tab)
- [x] Mouse hover selection with visual feedback
- [x] Embedded traffic light buttons
- [x] Tab filtering toggle
- [x] Filters tab loading spinner
- [x] Clean build with no warnings

## Notes

These changes are designed for users who prefer instant feedback and minimal latency. All new features have sensible defaults and are configurable via Settings.

### DIFF
--- a/DockDoor/Views/Hover Window/WindowPreviewHoverContainer.swift
+++ b/DockDoor/Views/Hover Window/WindowPreviewHoverContainer.swift
@@ -530,8 +530,6 @@ struct WindowPreviewHoverContainer: View {
             .globalPadding(20)
         }
         .padding(2)
-        .animation(nil, value: previewStateCoordinator.windows)
-        .animation(nil, value: previewStateCoordinator.currIndex)
         .onChange(of: previewStateCoordinator.currIndex) { newIndex in
             guard previewStateCoordinator.shouldScrollToIndex else { return }
             if showAnimations {


suatkocar:
Hi @ejbills,

Following up on my previous live preview contribution, this PR focuses on Window Switcher performance.

**Context:** I noticed some latency when cycling through windows - Tab key felt sluggish compared to Shift key, and mouse hover interactions weren't clearly distinguished from keyboard selection. The changes aim to provide a more responsive experience for users who prefer instant feedback.

**Key points:**
- All changes have sensible defaults
- New features are opt-in via Settings
- No breaking changes to existing functionality

Happy to make adjustments based on your feedback!

ejbills:
I think, in light of this PR unilaterally removing lots of varying animation values, I would appreciate a before/after showing the changes (in a video? possibly?) and how it benefits improving performance, and why the animation should not instead be gated behind the showAnimations toggle.

ejbills:
Also, additional features like 'embedding xyz in window switcher' should be scoped to their own PR.

ejbills:
Additionally, there's a critical bug in the Tab Filtering feature that needs to be addressed (which should also be its own PR).
The filterOutTabbedWindows function uses CGWindowListCopyWindowInfo([.optionOnScreenOnly, ...]) which only returns windows visible on the current space. This incorrectly filters out all windows from other desktops/spaces - not browser tabs.
Testing showed legitimate windows users expect to see, regardless of the "Show tabs as windows" setting, being hidden. The current implementation breaks multi-space workflows.


This feature would need a completely different approach to detect actual browser tabs (which is complex and browser-specific), or should be removed from this PR.

suatkocar:
Thanks for the detailed feedback. I'm restructuring the PR now:

**Regarding animations:** You're right, I've updated the hover animation to respect the existing `showAnimations` toggle instead of removing it entirely. The animation now only runs when `showAnimations` is enabled.

**Regarding separate PRs:** Agreed. I'm splitting this into focused PRs:
- This PR will focus only on **mouse hover selection + performance improvements**
- Embedded window switcher controls → separate PR
- Tab filtering → separate PR (and noted the critical bug you found with `optionOnScreenOnly` - that approach definitely won't work for detecting browser tabs)
- Other UX improvements → separate PR

I'll push the cleaned up version shortly.

PR #896, opened 2025-12-09
feat: add granular live preview settings and improve mouse hover selection

suatkocar:
## Summary

This PR enhances Window Switcher performance and user experience by introducing granular live preview controls and improving mouse hover selection behavior.

## Why These Changes Are Combined

These features were developed and tested together because they are **interdependent**:

1. **Live preview scope** (`hoveredWindowOnly`, `hoveredAppWindows`, `allWindows`) directly uses `hoveredIndex` and `focusedIndex` from the mouse hover system
2. **Testing live preview performance** requires mouse hover functionality to switch between windows
3. **Both features address the same core issue**: Window Switcher can become laggy when live preview is enabled, and users need tools to find their optimal balance

**Without live preview:** Window Switcher runs smoothly and responsively.
**With live preview at high settings:** Mouse hover and Tab cycling can become laggy.

By providing granular controls, users can configure settings based on their hardware capabilities.

---

### 🎬 Live Preview Enhancements

**Separate controls for Dock and Window Switcher:**

| Setting | Dock Default | Switcher Default |
|---------|--------------|------------------|
| Enabled | ✅ On | ❌ Off |
| Quality | High | Low |
| Frame Rate | 24 FPS | 10 FPS |

**Quality Options:** Thumbnail (320px), Low (480px), Standard (640px), High (960px), Retina (1280px), Native (Best)

**Frame Rate Options:** 5, 10, 15, 24, 30, 60, 120 FPS

**Window Switcher Live Preview Scope:**
| Option | Description | Performance Impact |
|--------|-------------|-------------------|
| Hovered Window Only | Only selected window gets live preview | Minimal |
| Hovered App's Windows | All windows from selected app get live preview | Moderate (default) |
| All Windows | Every window gets live preview | May cause lag with many windows |

---

### 🖱️ Mouse Hover Selection Improvements

**Problem from previous PR (#890):** Confusing dual selection states where both keyboard and mouse could highlight different windows simultaneously.

**New Behavior:**

| Scenario | Behavior |
|----------|----------|
| Switcher opens | Keyboard selection active, mouse hover disabled until movement detected |
| Mouse moves | Mouse hover becomes active, selection follows cursor |
| Mouse stops, Tab pressed | Keyboard takes over from current position |
| Mouse moves again | Mouse hover resumes |

**Key Improvements:**
- **Single selected window** - Only one window highlighted at a time, no confusion
- **Activity-based precedence** - Last input type (keyboard or mouse) determines active selection
- **Movement threshold detection** - Prevents accidental selection when switcher opens under cursor
- **Keyboard continuity** - When mouse stops, keyboard navigation continues from current position (index not reset)

---

### ⚡ Performance Optimizations

| Change | Before | After | Benefit |
|--------|--------|-------|---------|
| Hover transition animation | Always 0.175s | Respects `showAnimations` setting | Accessibility + snappier feel when disabled |
| `bringToFront` retry sleep | 50ms | 20ms | Faster window activation |
| Main thread safety | Not checked | AX operations check `Thread.isMainThread` | Prevents potential crashes |
| Switcher live preview | N/A | Disabled by default, Low quality, 10 FPS | Smooth out-of-box experience |

---

## Technical Notes

**Why `DispatchQueue.main.async` instead of `Task { @MainActor }`?**
- Lower overhead for Timer callbacks
- More predictable execution in event handling context

**Why NotificationCenter for mouse hover coordination?**
- Loose coupling between `KeybindHelper` and `PreviewStateCoordinator`
- Allows keyboard repeat timers to stop when mouse takes over without tight dependencies

**Why movement threshold of 0.001?**
- Detects any intentional mouse movement
- Prevents false activation when switcher appears under stationary cursor

---

## New Settings

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `enableMouseHoverInSwitcher` | Bool | `true` | Enable mouse hover selection in Window Switcher |
| `enableLivePreviewForDock` | Bool | `true` | Enable live preview for Dock hover |
| `enableLivePreviewForWindowSwitcher` | Bool | `false` | Enable live preview for Window Switcher |
| `dockLivePreviewQuality` | Enum | `.high` | Dock preview quality |
| `dockLivePreviewFrameRate` | Enum | `.fps24` | Dock preview frame rate |
| `windowSwitcherLivePreviewQuality` | Enum | `.low` | Switcher preview quality |
| `windowSwitcherLivePreviewFrameRate` | Enum | `.fps10` | Switcher preview frame rate |
| `windowSwitcherLivePreviewScope` | Enum | `.hoveredAppWindows` | Which windows get live preview |

---

### Live Preview
- [ ] Dock live preview with different quality/FPS settings
- [ ] Window Switcher live preview disabled (default) - should be smooth
- [ ] Window Switcher live preview enabled with each scope option
- [ ] Quality/FPS combinations at different window counts

### Mouse Hover
- [ ] Open switcher, don't move mouse → keyboard navigation works
- [ ] Move mouse → selection follows cursor
- [ ] Stop mouse, press Tab → keyboard continues from current position
- [ ] Rapid Tab/Shift+Tab cycling performance
- [ ] Hold Tab / Hold Shift+Tab for continuous cycling

### Performance
- [ ] `showAnimations` disabled → no hover transition animations
- [ ] Many windows open → switcher remains responsive with conservative settings

---

## Related

- Supersedes mouse hover changes from #890
- Addresses performance concerns and animation feedback from #890 review

---

## Request

🙏 Would appreciate a timely review to avoid additional merge conflicts from ongoing main branch activity. These changes have been extensively tested and are ready for integration.

### DIFF
--- a/DockDoor/Localizable.xcstrings
+++ b/DockDoor/Localizable.xcstrings
@@ -4787,6 +4787,18 @@
         }
       }
     },
+    "5 FPS (Battery Saver)" : {
+
+    },
+    "10 FPS" : {
+
+    },
+    "15 FPS" : {
+
+    },
+    "24 FPS (Film)" : {
+
+    },
     "30 FPS" : {
 
     },
@@ -7855,6 +7867,12 @@
     },
     "All Windows" : {
       "comment" : "Switcher invocation mode"
+    },
+    "All windows from the hovered app get live preview" : {
+
+    },
+    "All windows get live preview (may cause lag with many windows)" : {
+
     },
     "Allow dynamic image sizing" : {
       "localizations" : {
@@ -25804,6 +25822,9 @@
         }
       }
     },
+    "Dock Frame Rate" : {
+
+    },
     "Dock Preview Aero Shake Action" : {
       "localizations" : {
         "af" : {
@@ -27709,6 +27730,9 @@
           }
         }
       }
+    },
+    "Dock Quality" : {
+
     },
     "DockDoor does not record your screen or audio. It only captures static window previews. No information is stored or shared; all processing occurs privately on your device." : {
       "localizations" : {
@@ -29996,6 +30020,12 @@
     "Enable dock scroll gestures" : {
 
     },
+    "Enable for Dock Preview" : {
+
+    },
+    "Enable for Window Switcher" : {
+
+    },
     "Enable gestures in window switcher" : {
 
     },
@@ -30194,6 +30224,9 @@
       }
     },
     "Enable Live Preview (Video)" : {
+
+    },
+    "Enable mouse hover selection" : {
 
     },
     "Enable Pinning" : {
@@ -39768,10 +39801,10 @@
         }
       }
     },
-    "High" : {
+    "High (960px)" : {
 
     },
-    "Higher quality and frame rate use more CPU/GPU resources." : {
+    "Higher quality and frame rate use more CPU/GPU resources. Use lower settings for Window Switcher if you experience lag." : {
 
     },
     "Highlight Gradient Colors" : {
@@ -40348,6 +40381,12 @@
           }
         }
       }
+    },
+    "Hovered App's Windows" : {
+
+    },
+    "Hovered Window Only" : {
+
     },
     "How to Set Up" : {
       "extractionState" : "stale",
@@ -45890,12 +45929,6 @@
         }
       }
     },
-    "Live Preview Frame Rate" : {
-
-    },
-    "Live Preview Quality" : {
-
-    },
     "Loading lyrics..." : {
       "localizations" : {
         "af" : {
@@ -46465,6 +46498,9 @@
           }
         }
       }
+    },
+    "Low (480px)" : {
+
     },
     "Lower values make gestures more sensitive. Higher values require longer swipes. Applies to both dock previews and window switcher." : {
 
@@ -50106,6 +50142,9 @@
           }
         }
       }
+    },
+    "Native (Best)" : {
+
     },
     "Navigate to Settings → Privacy & Security" : {
       "localizations" : {
@@ -55248,6 +55287,9 @@
           }
         }
       }
+    },
+    "Only the window under the mouse gets live preview" : {
+
     },
     "Open Accessibility Settings" : {
       "extractionState" : "stale",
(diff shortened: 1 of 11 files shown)


ejbills:
This PR needs to be split. The live preview granular settings are fine, but the mouse hover changes need to be reworked entirely.

### You added global mouse polling while claiming performance improvements

```swift
mouseMovedMonitor = NSEvent.addLocalMonitorForEvents(matching: .mouseMoved) { [weak self] event in
    self?.handleMouseMoved()
    return event
}
```

`NSEvent.addLocalMonitorForEvents(.mouseMoved)` fires on **every single mouse movement** while the switcher is active. We already use SwiftUI's `.onHover` which only fires when entering/leaving a view. That's the performant approach - it's what SwiftUI is optimized for.

The "movement threshold" problem you're trying to solve? That's just a boolean check in the existing `onHover`:

```swift
.onHover { isHovering in
    if isHovering && !hasMovedSinceOpen {
        hasMovedSinceOpen = true
        return
    }
    // existing logic
}
```

No need to poll mouse coordinates globally.

### The cached state adds memory overhead, not performance

```swift
private var cachedWindows: [WindowInfo] = []
private var cachedMouseLocation: CGPoint = .zero
private var cachedTargetScreen: NSScreen?
```

Before: values passed directly to the async task.
After: stored as instance variables, keeping `WindowInfo` objects and their `CGImage` references alive longer.

This increases memory pressure for no benefit.

### State explosion

We had `currIndex`. Now we have:
- `focusedIndex`
- `hoveredIndex`  
- `currIndex` (computed, just returns `focusedIndex`)
- `pendingHoverIndex`
- `hasNotifiedMouseActivity`
- `isMouseHoverEnabled`
- `initialMousePosition`
- `lastActivityType`

Plus NotificationCenter coordination between classes. All of this could be handled by SwiftUI's `.onHover` + a single `lastInputWasKeyboard: Bool`.

There is a lot that needs to change:
1. Remove the `NSEvent.addLocalMonitorForEvents` approach - use the existing `.onHover` and work it into the existing architechture
2. Handle "ignore first hover" in the existing callback, not with global mouse tracking
3. Single flag for keyboard vs mouse, not parallel index systems
4. Drop the cached state
5. Drop the NotificationCenter coordination

Split the PR. Live preview settings can merge. Mouse hover needs a rewrite that works *with* SwiftUI's hover system instead of adding expensive polling on top of it.

Issue #822, opened 2025-11-09
[BUG] "Scroll to window on mouse hover" shouldn't activate if mouse hasn't moved


hexcrete:
## Describe the bug
The option of "Scroll to window on mouse hover" is very useful since it also selects the window under the mouse, enabling faster interaction across multiple windows, but it currently has an issue making it hard to use: 
- When the Window Switcher activates and this option is on, the window under the cursor gets immediately selected. 
- This makes it very inconvenient if the user just intended to quickly Alt+Tab to a specific window, but now has to reach for the mouse and position it at the correct window or use the arrow keys. 

## Steps to reproduce
1. Enable "Scroll to window on mouse hover" 
2. Activate Window Switcher using the shortcut, do not touch the mouse/trackpad 
3. If the cursor has been positioned over any window in the Window Switcher, the window will get selected 
5. Releasing initializer key will activate the window under the cursor, even if it hasn't moved since before the Window Switcher activation 

## Expected behavior
1. Enable "Scroll to window on mouse hover" 
2. Activate Window Switcher using the shortcut, do not touch the mouse/trackpad 
3. The previous active window will be selected 
4. If you move the cursor now, the window under the cursor will be selected 

## Actual behavior
See "Steps to reproduce"

## Environment
 - macOS version: macOS 26.1
 - App version: 1.25.5

## Additional context
- It might be a good idea to introduce a movement threshold for this. Some users (niche) might use a mouse so sensitive even a small table bump might move the cursor and inadvertently activate the wrong window. 
- The "Scroll to window on mouse hover" sells this feature short. It is extremely useful to select the window under the cursor when moved. It is less useful to auto scroll the view. I suggest renaming it ("Select window on mouse hover") and potentially even let the user disable the auto-scroll and only keep the auto-select. 

- [x] I have reviewed the open issues to ensure this is not a duplicate

PR #901, opened 2025-12-10
refactor: improve mouse hover selection in Window Switcher

suatkocar:
https://github.com/user-attachments/assets/8c8dc8ed-322f-4703-b55f-4d555868d3c4

## Summary

This PR refactors the mouse hover selection feature in Window Switcher, addressing all feedback from #896.

### Global mouse polling → Removed

**Before:** `NSEvent.addLocalMonitorForEvents(.mouseMoved)` 

After: SwiftUI's `.onContinuousHover` - provides cursor location on each hover event (no global polling)

### Cached state adds memory overhead → Removed

**Before:** `cachedWindows`, `cachedMouseLocation`, `cachedTargetScreen`

**After:** No cached state

### State explosion → Minimal state

**Before (8+ variables):** `focusedIndex`, `hoveredIndex`, `currIndex`, `pendingHoverIndex`, `hasNotifiedMouseActivity`, `isMouseHoverEnabled`, `initialMousePosition`, `lastActivityType`

**After (3 new variables):** `hasMovedSinceOpen`, `lastInputWasKeyboard`, `initialHoverLocation`

### NotificationCenter coordination → Removed

**After:** Direct `@Published` property read in timer callbacks

### Handle ignore first hover in existing callback → Done

Movement threshold check happens inside `.onContinuousHover` callback.

### Hover Detection
- Use `.onContinuousHover` instead of `.onHover` for location tracking
- 1px movement threshold prevents accidental selection on switcher open
- Respect Reduce Motion setting for hover animations

### Keyboard/Mouse Coordination
- `lastInputWasKeyboard` tracks input source for seamless transitions
- Keyboard continues from current position when mouse stops
- `isKeyboardScrolling` prevents hover during keyboard scroll
- Timer-based Tab/Shift repeat stops when mouse takes over

### Settings
- Rename `scrollToMouseHoverInSwitcher` → `enableMouseHoverInSwitcher`
- Updated toggle label and description for clarity

## Architecture

| Concern | #896 | This PR |
|---------|------|---------|
| Mouse tracking | `NSEvent.addLocalMonitorForEvents` | `.onContinuousHover` |
| Cached state | 3 variables | None |
| New state | 8+ variables | 3 variables |
| Coordination | `NotificationCenter` | Direct property read |
| Index system | Parallel indexes | Single `currIndex` |

## Testing

- [ ] Open switcher, don't move mouse → Tab works
- [ ] Move mouse → selection follows cursor
- [ ] Stop mouse, press Tab → continues from current position
- [ ] Hold Tab + move mouse → timer stops, mouse takes over
- [ ] Reduce Motion enabled → no hover animations
- [ ] 2-row/4-row layouts with scroll

### DIFF
--- a/DockDoor/Utilities/KeybindHelper.swift
+++ b/DockDoor/Utilities/KeybindHelper.swift
@@ -40,11 +40,21 @@ private class WindowSwitchingCoordinator {
         }
 
         if stateManager.isActive {
+            // TODO: Consolidate WindowSwitcherStateManager and PreviewStateCoordinator into a single index system
+            let uiIndex = previewCoordinator.windowSwitcherCoordinator.currIndex
+            if uiIndex >= 0, uiIndex != stateManager.currentIndex {
+                stateManager.setIndex(uiIndex)
+            }
+
             if isShiftPressed {
                 stateManager.cycleBackward()
             } else {
                 stateManager.cycleForward()
             }
+
+            previewCoordinator.windowSwitcherCoordinator.hasMovedSinceOpen = false
+            previewCoordinator.windowSwitcherCoordinator.initialHoverLocation = nil
+
             previewCoordinator.windowSwitcherCoordinator.setIndex(to: stateManager.currentIndex)
         } else if isModifierPressed {
             await initializeWindowSwitching(
--- a/DockDoor/Views/Hover Window/Shared Components/SharedPreviewWindowCoordinator.swift
+++ b/DockDoor/Views/Hover Window/Shared Components/SharedPreviewWindowCoordinator.swift
@@ -576,6 +576,9 @@ final class SharedPreviewWindowCoordinator: NSPanel {
         let coordinator = windowSwitcherCoordinator
         guard !coordinator.windows.isEmpty else { return }
 
+        coordinator.hasMovedSinceOpen = false
+        coordinator.initialHoverLocation = nil
+
         let threshold = Defaults[.windowSwitcherCompactThreshold]
         let isListViewMode = coordinator.windowSwitcherActive && threshold > 0 && coordinator.windows.count >= threshold
 
--- a/DockDoor/Views/Hover Window/WindowPreview Supporting/PreviewStateCoordinator.swift
+++ b/DockDoor/Views/Hover Window/WindowPreview Supporting/PreviewStateCoordinator.swift
@@ -5,6 +5,9 @@ import SwiftUI
 class PreviewStateCoordinator: ObservableObject {
     @Published var currIndex: Int = -1
     @Published var windowSwitcherActive: Bool = false
+
+    @Published var hasMovedSinceOpen: Bool = false
+    var initialHoverLocation: CGPoint?
     @Published var fullWindowPreviewActive: Bool = false
     @Published var windows: [WindowInfo] = []
     @Published var shouldScrollToIndex: Bool = true
@@ -45,6 +48,11 @@ class PreviewStateCoordinator: ObservableObject {
             fullWindowPreviewActive = toState
         case .none:
             return
+        }
+
+        if !oldSwitcherState, windowSwitcherActive {
+            hasMovedSinceOpen = false
+            initialHoverLocation = nil
         }
 
         // If window switcher state changed and we have windows, recalculate dimensions
--- a/DockDoor/Views/Hover Window/WindowPreview.swift
+++ b/DockDoor/Views/Hover Window/WindowPreview.swift
@@ -16,7 +16,7 @@ struct WindowPreview: View {
     let dimensions: WindowPreviewHoverContainer.WindowDimensions
     let showAppIconOnly: Bool
     let mockPreviewActive: Bool
-    let onHoverIndexChange: ((Int?) -> Void)?
+    let onHoverIndexChange: ((Int?, CGPoint?) -> Void)?
     var isEligibleForLivePreview: Bool = true
 
     @Default(.windowTitlePosition) var windowTitlePosition
@@ -48,6 +48,7 @@ struct WindowPreview: View {
     @Default(.dockLivePreviewFrameRate) var dockLivePreviewFrameRate
     @Default(.windowSwitcherLivePreviewQuality) var windowSwitcherLivePreviewQuality
     @Default(.windowSwitcherLivePreviewFrameRate) var windowSwitcherLivePreviewFrameRate
+    @Default(.showAnimations) var showAnimations
 
     @State private var isHoveringOverDockPeekPreview = false
     @State private var isHoveringOverWindowSwitcherPreview = false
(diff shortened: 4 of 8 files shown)


suatkocar:
## Design Decisions

### Why `.onContinuousHover` instead of `.onHover`?

Two reasons:

1. **Location tracking**: `.onHover` only provides enter/leave events. We need cursor location for the 1px movement threshold to detect intentional movement vs. switcher opening under a stationary cursor.

2. **Better UX**: With `.onHover`, if the user hovers over Window A, then uses keyboard to navigate to Window B, they cannot re-select Window A with mouse without leaving and re-entering it. `.onContinuousHover` allows immediate selection of the window under cursor after keyboard navigation - no awkward "exit and re-enter" required.

**Performance note**: To prevent excessive re-renders from continuous events, we guard state updates with `if !isHovering` checks - state only updates when it actually changes.

**Note on `.onContinuousHover`**: This is a SwiftUI-native view modifier, not global polling. Events are delivered only when the cursor is over the specific view - no system-wide monitoring.

### Why 1px movement threshold?

When the switcher opens, the cursor might already be over a window. Without a threshold, this would immediately select that window, overriding the keyboard selection. The 1px threshold ensures only intentional mouse movement triggers hover selection. Once movement is detected and `hasMovedSinceOpen` becomes `true`, the threshold check is skipped entirely on subsequent hover events to avoid unnecessary state updates.

### Why unified selection instead of separate keyboard/mouse indexes?

**Before (main branch)**: When mouse hover selection was enabled, keyboard and mouse had separate selection states. This caused several UX issues:

1. **Visual confusion**: Two different windows could appear "selected" simultaneously - one by keyboard, one by mouse hover. With active window border enabled, the screen became cluttered.

2. **Ambiguous "real" selection**: Users couldn't tell which selection would actually activate on Enter/click - the keyboard selection or the hover selection.

3. **Inconsistent continuation**: Keyboard navigation would continue from its own last position, ignoring where mouse left off, making the interaction feel disconnected.

**After (this PR)**: Single `currIndex` that both keyboard and mouse update. Whichever input was used last determines the current selection. When you stop using mouse and press Tab, it continues from where the mouse left off - intuitive and predictable.

### Why `isKeyboardScrolling` flag with 0.3s delay?

In 2-row layouts, keyboard navigation triggers scroll animation (`.snappy` ~0.3s). During this animation, cursor position relative to windows changes, which would trigger hover events and conflict with keyboard selection. The flag blocks hover during the scroll animation duration.

### Why Timer-based Tab/Shift cycling?

Tab cycling existed before but was slower and handled differently. We unified and improved the cycling logic:

- **Tab repeat**: Previously existed but was processed through `handleWindowSwitching` which added overhead. Now uses direct timer for faster response.
- **Shift cycling**: Requires holding the modifier key (e.g., Cmd+Shift). In practice, holding Cmd while also pressing Tab repeatedly is awkward, so users prefer holding Shift to cycle backward continuously.
- **Unified timing**: Both now use the same timer pattern - 0.4s initial delay, 0.05s repeat interval - matching system keyboard repeat behavior for consistent feel.

### Why check `lastInputWasKeyboard` in timer callback?

This replaces NotificationCenter coordination. When mouse hover changes the index, it sets `lastInputWasKeyboard = false`. The timer callback checks this flag and stops itself if mouse took over - clean coordination without notification overhead.

### Setting rename: `scrollToMouseHoverInSwitcher` → `enableMouseHoverInSwitcher`

The original setting name was misleading - it controlled both hover selection AND scroll behavior. The new name clearly indicates it enables/disables the entire mouse hover selection feature. Default remains `false` (disabled).

ejbills:
Please provide a human-written summary of this PR.

suatkocar:
So basically i rewrote the whole mouse hover selection thing from scratch based on ur feedback. Ditched the NSEvent polling stuff and switched to SwiftUI's onContinuousHover which is way cleaner. Also removed all those cached state variables that were causing the memory overhead u mentioned.

The main thing now is keyboard and mouse both use the same currIndex - so theres no more weird situation where two windows look selected at the same time. Whatever input u used last is what counts.

And I added a small 1px movement threshold so when the switcher opens it doesnt immediately select whatever window happens to be under ur cursor. Only intentional mouse movement triggers selection now.

suatkocar:
This PR closes the following issue:

- Closes #822 - "Scroll to window on mouse hover" shouldn't activate if mouse hasn't moved → 1px movement threshold added to prevent accidental selection when switcher opens under stationary cursor

suatkocar:
https://github.com/user-attachments/assets/abc50e5a-b2bb-4396-897e-8bbda0e68bee

Added a separate toggle for scroll behavior - users can now enable mouse hover selection without auto-scroll.

suatkocar:
@ejbills I noticed that significant portions of my implementation were removed during the merge, and the current version has performance and functionality issues.

### Removed: Timer-based Tab/Shift Cycling

My implementation used direct timer-based cycling with 0.05s repeat interval. The current version routes every single Tab press through `handleWindowSwitching()`.

**Overhead per Tab press in current version:**

| Check | Purpose | Necessary for Tab? |
|-------|---------|-------------------|
| `async` dispatch | Non-blocking call | No - can be done directly on UI thread |
| `isProcessingSwitcher` guard | Prevent double calls | No - timer controls this |
| `stateManager.isActive` check | Is switcher open? | No - already open when Tab pressed |
| `previewCoordinator.isVisible` check | Is it visible? | No - redundant with above |
| Index sync between two systems | Sync `stateManager` ↔ `currIndex` | Design issue - should be single index |
| State resets on every Tab | Clear mouse state | No - just need to know input source |
| `setIndex()` call | Update selection | Yes - but only this is actually needed |

My version was a direct `cycleForward()` + `setIndex()` call - no async, no guards, no state resets. The difference is noticeable even on single Tab presses, not just when holding.

Yes, my timer-based implementation had more code - but you removed it only to add a similar amount of code for edge scroll. The net result is roughly the same code volume, except now there's noticeable lag on single Tab press and cycling that was the main reason of my PR.

### Removed: Shift-Hold Backward Cycling

My implementation had `shiftCycleTimer` - users could press ONLY Shift and/or hold Shift to cycle backward continuously. The current version requires pressing Cmd+Shift+Tab repeatedly for each backward step. This is a significant UX regression.

### Removed: Keyboard/Mouse Coordination

`lastInputWasKeyboard` and `isKeyboardScrolling` were removed. Without these:
- Hover events can conflict during keyboard scroll animation (0.3s `.snappy` animation)
- No tracking of which input source was used last for seamless transitions
- Timer callbacks can't detect when mouse takes over to stop themselves

### Removed: Separate Scroll Toggle

It seems this was removed to make way for edge scroll, but it was a useful feature that could have coexisted. `scrollOnMouseHoverInSwitcher` allowed users to enable hover selection without auto-scroll. Removing it in favor of edge scroll wasn't necessary - both features could work together.

### Edge Scroll Issues

The added edge scroll feature has several issues:

1. **Fixed speed** - hardcoded `4.0 px`, no user configuration (better on 16.0 px)
2. **60fps timer with view traversal** - `findScrollView(in:)` recursively walks the entire NSView hierarchy on every single frame
3. **No dynamic acceleration** - same speed regardless of cursor position relative to edge
4. **Window edges only** - doesn't work when cursor reaches screen boundaries
5. **Uses `.onHover`** - only gets enter/leave events, no continuous position tracking

I've addressed all of these in my `local-usage` branch:
- Configurable speed slider (4-48px)
- `dynamicEdgeScrollSpeed` with `calculateSpeedMultiplier()` based on edge distance
- `screenEdgeMonitor` for screen boundary support
- Separate `enableEdgeScrollInSwitcher` toggle
- Added granular scroll direction controls (`scrollHorizontallyOnHover`, `scrollVerticallyOnHover`)

The scroll direction controls came from noticing that the switcher has a row limit setting but no column limit - users often complain about the layout algorithm, and having granular scroll control helps with different layout configurations.

Feel free to check it out if you're interested.

PR #938, opened 2025-12-19
fix: window names not being updated when restarting an app (ex vs code) and fix screen recording permission pop up appearing over and over

maddada:
## Describe your changes

1- Fixed issue where video permission pop up kept appearing every few seconds if the user didn't provide it (happens since last version only) 

2- Fixes this issue where all of the windows have the same name after restarting visual studio code.
Had to restart DockDoor to fix this.

2025-12-19_Code - Insiders_01-41-57

2025-12-19_Code - Insiders_01-57-10

### DIFF
--- a/DockDoor/Components/PermissionsView/PermissionsChecker.swift
+++ b/DockDoor/Components/PermissionsView/PermissionsChecker.swift
@@ -6,9 +6,7 @@ class PermissionsChecker: ObservableObject {
     @Published var screenRecordingPermission: Bool = false
     private var timer: AnyCancellable?
 
-    // Cached permission value - updated by timer, read by WindowUtil
-    private static var cachedScreenRecordingPermission: Bool = // Initialize with actual check on first access
-        checkScreenRecordingPermissionFromCG()
+    private static var cachedScreenRecordingPermission: Bool = checkScreenRecordingPermissionFromCG()
 
     init() {
         checkPermissions()
@@ -34,12 +32,10 @@ class PermissionsChecker: ObservableObject {
         return result
     }
 
-    /// Returns the cached screen recording permission value (cheap read)
     static func hasScreenRecordingPermission() -> Bool {
         cachedScreenRecordingPermission
     }
 
-    /// Performs the actual CG API check (expensive)
     private static func checkScreenRecordingPermissionFromCG() -> Bool {
         let stream = CGDisplayStream(
             dispatchQueueDisplay: CGMainDisplayID(),


ejbills:
Did you determine that vs code doesn't send any other notification, besides the title change? A lot of notifications cover the same use case, so its possible something like kAXWimdowChanged or something could fire on title change which we also listen for.

ejbills:
In the updateAllWindowsIn... func we shouldn't guard these based on permissions. This method interfaces either the AX-discovered windows (which is how it still shows windows without screen permissions). With this added, the AX references will remain stale. 

Look at the end of this func

maddada:
> In the updateAllWindowsIn... func we shouldn't guard these based on permissions. This method interfaces either the AX-discovered windows (which is how it still shows windows without screen permissions). With this added, the AX references will remain stale. 
> 
> Look at the end of this func

Sorry this is an unrelated fix for getting the permission request for video over and over if you denied it before. (Not being able to use the app with this permission disabled)

I'll remove the guard here and check if the permission request keeps happening and if window titles don't get affected.

maddada:
> Did you determine that vs code doesn't send any other notification, besides the title change? A lot of notifications cover the same use case, so its possible something like kAXWimdowChanged or something could fire on title change which we also listen for.

Will check more and get back to you. I just solved it this way without looking at the existing notifications well first.

ejbills:
The pattern you used in `updateNewWindowsForApp()` is correct - wrap only the SCK block in the permission check, but let the AX fallback code after it run unconditionally. Apply the same pattern to `updateAllWindowsInCurrentSpace()`.

The permission triggering APIs that need guards are:
- `SCShareableContent.excludingDesktopWindows()` 
- `SCStream.startCapture()`
- `CGSHWCaptureWindowList()`

You will need to validate if CGWindowListCopyWindowInfo triggers the permission box. AX-based window discovery works without screen permission and should remain unguarded.

maddada:
Implemented the change to the guard you asked for.

Also I checked and couldn't find another notification that can be used to update the title for vs code

Logs are here:
[logs.md](https://github.com/user-attachments/files/24280835/logs.md)

maddada:
Summary by Claude Code:

1. Window Title Tracking for VS Code (WindowManipulationObservers.swift)

  Problem: VS Code window titles weren't updating after restarting.

  Solution: Added kAXTitleChangedNotification observer to track window title changes.
```
  // Added to observer registration (3 places: deinit, createObserverForApp, removeObserver)
  AXObserverAddNotification(observer, appElement, kAXTitleChangedNotification as CFString, ...)

  // Handler for title changes
  case kAXTitleChangedNotification:
      // Only process if the element is actually a window
      guard let role = try? element.role(), role == kAXWindowRole as String else { return }
      let windowID = try? element.cgWindowId()
      handleWindowEvent(...) { windowSet in
          update(windowSet: &windowSet, matching: windowID, element: element) { window in
              if let freshTitle = try? window.axElement.title(), !freshTitle.isEmpty {
                  window.windowName = freshTitle
              }
          }
      }
```
  ---
  2. Screen Recording Permission Checks (WindowUtil.swift, LiveWindowCapture.swift)

  Problem: Screen recording permission popup kept appearing repeatedly.

  Solution: Added permission checks before calling SCShareableContent and CGSHWCaptureWindowList APIs.

  - Added hasScreenRecordingPermission() helper function
  - Guards around all SCShareableContent calls
  - Guard around captureWindowImage
  - Added AX fallback for window discovery when permission not granted

  ---
  3. Added Restart Menu Item (AppDelegate.swift) to speed up testing
  Added "Restart DockDoor" option to the menu bar menu.

ejbills:
hasScreenRecordingPermission calls the CG API which can be computationally expensive. Can we run this check once at app startup in the AppDelegate? That way, we can cache the value and then just read the stored value, rather than running hasScreenRecordingPermission over and over, for no reason. 

I will add this when I can get to it

maddada:
Done, pushed a commit that addressed you last comment.

Changes done:
The expensive CG API check now only runs once at app startup then the cached value is used afterwards. All other permission checks throughout the codebase now just read the cached boolean value.

On a separate note, JFYI the existing code has polling every second for a CG api in settings if the user is on the support page, but I think this doesn't make sense (summary by AI below):

- The PermissionsChecker class has a timer that polls every 1 second to detect permission changes while the settings view is open
- The intent is to update the UI reactively when the user grants/revokes screen recording permission in System Preferences
- However, this polling is unnecessary because macOS requires an app restart for screen recording permission changes to take effect
- macOS itself displays a message stating "quit and reopen [app] to grant access" when enabling screen recording
- Therefore, checking every second provides no benefit - the cached value from app startup will remain accurate for the entire session
- The expensive CGDisplayStream API call runs every second when settings is open, wasting CPU cycles for a permission that cannot change mid-session
- Recommendation: The timer-based polling for screen recording permission could be removed entirely, keeping only the initial check at app startup

Let me know what you think

PR #940, opened 2025-12-19
feat: added sliders that let the user adjust the position of the window switcher (useful for large screens) and added toggle to align the switcher to the top instead of center it

maddada:
## Describe your changes

I really needed this feature because I use a 55" TV as my main monitor and I like to adjust the placement of the window switcher list to be lower since I use the lower 2/3 part of the screen mostly (top is for extra windows like youtube/whatsapp etc). Otherwise if I don't have this then I have to look up to see the recent windows.

I also added a toggle to make the window switcher start at the same position from the top always instead of changing vertical position based on how many windows there are.

The default behavior stays the same (aligned to center of the screen)

Added it to the bottom of advanced in the appearance tab

2025-12-19_DockDoor_10-42-57


ejbills:
consider consolidating all window switcher placement settings in one location - right now screen selection (placementStrategy, pinnedScreenIdentifier) is in MainSettingsView while the new offset/anchor settings are in AppearanceSettingsView. they all relate to where the switcher appears so grouping them together would make more sense

ejbills:
I found a bug, the offset is still being applied even when I have the offset position box unchecked.

PR #941, opened 2025-12-19
feat: added ability to make an app show up as just 1 entry in the window switcher to the filters page (still shows as multiple in active app switcher)

maddada:
## Describe your changes

I needed this feature because I have lots of VS Code windows open at all times (to switch between projects easily)

The issue that I face is that dock door starts showing 6 vs code entries and it makes the compact list view too long

With this mode then it shows only the most recent entry and hides the others when I cmd + tab normally

But when I use cmd + ` (which is my alternate mode key) which shows windows from the active app only, then it shows all of the windows of vs code and i can switch between them there.

Hope it's clear but please let me know if you need more explanation

https://github.com/user-attachments/assets/13d5e456-8945-4398-acc0-44edbc493913


ejbills:
Code looks good... I am just concerned about the placement of the feature in the settings. Other features that affect the sorting are in the general tab, but this requires the list of applications which ofc is only loaded in this filters tab. What do you think? @maddada

maddada:
@ejbills I think it's fine here since, would look out of place on the other page.
Up to you 😄

ejbills:
Let's put the checkbox to enbale this in the main settings with all of the other features, and then below the feature checkbox we can have a link titled 'configure groups' or something. Clicking this link will take the user to the list of apps in the other tab. We should be careful that the grouping section is not rendered in the filters tab if the option is enabled, to prevent confusion for the users who don't use it.

OR
I would support a change to the settings:
- Application list is moved to a component that is revealed with a sheet (refer to adding a word to the word filter, revealed by sheet here)
- We can call this shared component rather than rendering 2 application lists

maddada:
@ejbills I'm really sorry I didn't understand exactly how you want this to be implemented.
Can you please implement the one you think is better between these 2? I'll be happy to test it thoroughly after if you want.

ejbills:
@maddada ready for testing

maddada:
Will test it now and get back to you thanks. Was working on the other pr.

maddada:
Works awesome thank you!

PR #954, opened 2025-12-22
feat: restore the last recency order of windows when quitting/restarting

maddada:
## Describe your changes

Makes restarting/quitting restore the order of the windows in the window switcher as it was before quitting.

Supports "Recently used" and "Creation order (fixed)" window sort orders

Won't be able to restore ALL cases but it restores in most of them (only issue is if the window title changes between quitting and starting the app then it won't be able to be restored, but this is a bit rare so it's acceptable IMO)

Also if the app crashes then we won't save the last state, this is fine since the app has been very stable and almost never crashes (having a timer save the state every x seconds isn't good. It's fine to lose the order in case of crashing).

### DIFF
--- a/DockDoor/Utilities/Window Management/WindowOrderPersistence.swift
+++ b/DockDoor/Utilities/Window Management/WindowOrderPersistence.swift
@@ -19,9 +19,14 @@ enum WindowOrderPersistence {
         if cache == nil {
             let entries = Defaults[.persistedWindowOrder]
             cache = Dictionary(entries.map { ($0.key, $0) }, uniquingKeysWith: { first, _ in first })
+            DebugLogger.log("WindowOrderPersistence", details: "Loaded \(entries.count) persisted entries")
         }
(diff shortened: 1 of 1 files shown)


ejbills:
Will review when I get a chance, thanks for the PRs!

maddada:
No worries man, happy holidays and the thank you for the insanely useful and stable app. You're a legend!

DockDoor/Utilities/Window Management/WindowUtil.swift
```diff
@@ -248,6 +248,10 @@ enum WindowUtil {
 // MARK: - Cache Management
 
 extension WindowUtil {
+    static func getAllWindowsFromCache() -> [WindowInfo] {
```
ejbills: This is duplicating the getAllWindowsOfAllApps method, is there a reason this is a new method rather than being integrated into the existing methods?

PR #957, opened 2025-12-22
feat: Add support for parallel title and controls in window previews

X1nto:
## Describe your changes
This adds 4 new title and traffic controls alignment options, allowing you to align them at the same edges:

|                          | Left | Right |
|--------------------------|--------------------------|--------------------------|
| **Title top** | image | image|
| **Title bottom** | image | image |

P.S: Apart from introducing changes to the localization files due to these new alignment options, it looks like Xcode also added entries for a whole lot of new languages. Let me know if this needs to be addressed.

## Core Functionality Changes

### DIFF
--- a/DockDoor/Localizable.xcstrings
+++ b/DockDoor/Localizable.xcstrings
@@ -105,13 +105,13 @@
             "value" : ""
           }
         },
-        "nl" : {
+        "nb" : {
           "stringUnit" : {
             "state" : "new",
             "value" : ""
           }
         },
-        "no" : {
+        "nl" : {
           "stringUnit" : {
             "state" : "new",
             "value" : ""
@@ -295,13 +295,13 @@
             "value" : " "
           }
         },
-        "nl" : {
+        "nb" : {
           "stringUnit" : {
             "state" : "new",
             "value" : " "
           }
         },
-        "no" : {
+        "nl" : {
           "stringUnit" : {
             "state" : "new",
             "value" : " "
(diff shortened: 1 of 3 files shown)


ejbills:
> Apart from introducing changes to the localization files due to these new alignment options, it looks like Xcode also added entries for a whole lot of new languages. Let me know if this needs to be addressed.

You can safely remove the Localizable file from your commit history. It should only be showing the additions for the newly added strings.

ejbills:
Thanks for the PR - the feature itself is solid!

The current code already has some duplication with the enum flowing into the isLeadingControls mapping in both `windowSwitcherContent` and `dockPreviewContent`, which isn't great. But this PR makes it worse by moving that logic to even more call sites, resulting in those big if/else chains repeated throughout. This will be harder to maintain in the future if we want to make changes to the appearance.

Could you look at keeping this logic more encapsulated? Maybe a computed property on the enum or something similar - open to ideas. Just want to avoid scattering that mapping across a bunch of call sites. This will improve the call sites for all of the related options.

X1nto:
> The current code already has some duplication with the enum flowing into the isLeadingControls mapping in both `windowSwitcherContent` and `dockPreviewContent`, which isn't great. But this PR makes it worse by moving that logic to even more call sites, resulting in those big if/else chains repeated throughout. This will be harder to maintain in the future if we want to make changes to the appearance.
> 
> 
> 
> Could you look at keeping this logic more encapsulated? Maybe a computed property on the enum or something similar - open to ideas. Just want to avoid scattering that mapping across a bunch of call sites. This will improve the call sites for all of the related options.
> 
> 

Oh yeah, I was planning to use computed properties, but didn't want to change code drastically. I'll add an extension on WindowSwitcherControlPosition inside WindowPreview to avoid polluting WindowSwitcherControlPosition.

X1nto:
One small issue is that the value for isLeadingControls in windowSwitcherPreview and dockSwitcherPreview needs to be mirrored between top and bottom calls because of how the preview is structured internally. Might need to change that to actually make use of properties.

ejbills:
Is it cool if I commit to this branch?

X1nto:
> Is it cool if I commit to this branch?

Of course

ejbills:
Merged in  #977

Issue #451, opened 2025-01-09
[FR] Scroll on title bar window management


sm131313:
Scroll up on title bar should maximize or scroll down to minimize windows
similar to Swish (https://highlyopinionated.co/swish/)

2-finger swipe on title bar to move window to another desktop

2-finger swipe on window titles to pane or fullscreen windows

Vendrel:
I wanted to suggest the same thing, plus that the Hide/Unhide function could be activated if we do the same on the Dock Icon.

HyperDock does exactly all these, but development has stalled, it's very buggy, and it doesn't work on newer macOS versions.

It would result a seamless and obvious management of visible and hidden/minimized application windows, raising UX.

ejbills:
Can someone give me an example of this functionality in a video from hyperdock?

almt:
> Can someone give me an example of this functionality in a video from hyperdock?

@ejbills , see the first few seconds from [this Swish demo](https://highlyopinionated.co/swish/media/screens.mp4). Both Swish and Hyperdock supports a 2-finger swipe on the title bar to move a window to another desktop. But I believe Swish also requires a modifier key (which I wouldn't want)

ejbills:
This was added in #1141

almt:
> This was added in [#1141](https://github.com/ejbills/DockDoor/pull/1141)

@ejbills , believe #1141 only supports window sizing. But it doesn't support moving the window to another desktop?

Vendrel:
Just tested again, the problem still exists.

- Window is on `Space 2, Display 1`.
- Active Space: `Space 1, Display 1`.
- We can drag the Preview Panel to `Display 2`. ✓ 
- But if we drag the Preview Panel to `Space 1, Display 1`, it brings us to `Space 2, Display 2`.

ejbills:
Switching desktop spaces of windows is not supported as of right now, and should be tracked in another issue. This issue was closed in regards to executing window actions on the title bar scroll gesture. See #855

Vendrel:
Wow, that's one turn! :) But Dock can send windows to displays. Isn't it accessible? 

→ Send → Position → Job Done

PR #1141, opened 2026-03-08
feat: resize windows by scrolling on the title bar

lasncc2:
## Describe your changes

  This PR makes the title bar scroll gesture behavior more configurable and adds a quick restore flow for repeated vertical scroll actions.

  ### What changed
  - Added configurable settings for the centered window size when using the title bar scroll gesture
  - Added a configurable restore time window for repeated vertical scroll actions
  - Added restore behavior so repeating the same up/down title bar scroll action within the configured interval restores the previous window
  size and position
  - Updated the title bar scroll gesture settings UI and related localized strings

  ### Why
  The previous behavior used a fixed centered window size and applied maximize/center actions without an easy way to restore the original
  frame. This change gives users more control over the gesture behavior and makes repeated vertical scroll actions feel more predictable and
  reversible.

https://github.com/ejbills/DockDoor/issues/451

  ## AI Assistance Disclosure

  > **Policy**: AI-assisted contributions are welcome, but you must understand, test, and take full responsibility for your code. Pasting
  AI-generated code or PR descriptions without genuine human review and understanding will result in your PR being closed and may result in
  a ban from contributing.

  - [ ] No AI tools were used

  Used OpenAI Codex/ChatGPT to help summarize the code changes and draft the PR description based on the actual diff. I reviewed the
  generated text, verified that it matches my changes, and I take full responsibility for the final content and code in this PR.

  ## Core Functionality Changes

  This PR changes the behavior of title bar scroll gestures for focused windows:

  - Scrolling up still performs maximize, and scrolling down still performs center, but the centered window size is now configurable
  - Repeating the same vertical scroll action within a configurable time interval restores the window to its previous frame
  - The settings screen now exposes both the centered window size and the restore interval so users can tune the gesture behavior


ejbills:
Thank you for the PR, I will review it this weekend when I have time to test it out. 

Cool feature!

ejbills (changes requested):
Hey, thanks for the contribution. Cool idea but some things need to change before this can go in.

The main issue is the separate event tap, DockObserver already has a global event tap that handles scroll wheel events. Creating a second one means every scroll on the system hits two callbacks which is wasteful. This should just be integrated into DockObserver's existing callback and this also means TitleBarScrollObserver and its initialization in AppDelegate can go away.

On that note, the initialization doesn't follow the pattern we use. Every other observer in AppDelegate is gated on its feature toggle (see how DockObserver, KeybindHelper, etc are set up). Right now TitleBarScrollObserver gets created unconditionally even when the feature is off.

Some other things:
- `forceSingleRowForSideDock` defaults to `true` which changes behavior for existing users with left/right docks. Should default to `false`
- `cocoaFrame(for:)` duplicates the AX to Cocoa coordinate conversion that `WindowInfo.currentWindowFrame()` (which you also add in this PR) already does we can reuse that
- Chinese comments in DockUtils.swift and consts.swift, codebase is English
- `preventPreviewReentryDuringFadeOut` default silently changed from `false` to `true`, this should be called out if intentional
- In Window Image Sizing Calculations.swift you still inline the horizontal flow check in two places instead of using the `previewIsHorizontalFlow()` method you added

ejbills (approved):

PR #1333, opened 2026-05-10
feat: add dock click window restore setting

### LINKED ISSUE #1324: [FR] To not open all the window from the dock
## What problem does this solve?
When I click to open an app from the dock that has multiple windows, some of them are minimized, but it opens all the windows. The minize too.

Yes, I know that there is the option in the settings "Hide all app windows on dock icon click." And when I uncheck that, this behavior disappeared. But there should be an option that only minimizes the open windows, but when I click to open them back up, not all the windows should be open

## Proposed solution
To use how macOS opens a window from the dock. To open the last one that was active. 
And if I want to open a different one, I have the preview of the other window when I hover over the app icon on the dock (like now) 
Because opening all the windows, regardless of whether they were minimized, creates more clutter than it solves

## Additional context (optional)

- [x] I have reviewed existing issues to ensure this is not a duplicate

wiskiiii:
## Describe your changes

When the "Hide" option is checked, if the user clicks on an app in the dock that has more than one window, then all windows will be opened. This can make the screen a little messy. This PR keeps the behavior enabled by default, but adds an opt-out setting, so users can choos whether to open all windows when the icon is clicked.

Fixes #1324

- [ ] I have tested this PR on my own machine

## AI Assistance Disclosure

- [ ] No AI tools were used

I used Claude Code to implement the small setting change. I checked the diff and understood what these changes were doing, and confirmed that they were minor changes and not breaking ones

## Core Functionality Changes

This PR modifies Dock icon click behavior when "Hide all app windows on dock icon click" is enabled.

By default, behavior is unchanged: DockDoor still restores all minimized windows when reopening an app from the Dock. If the new "Restore all minimized windows on dock icon click" setting is disabled, DockDoor no longer batch-restores minimized windows during the Dock icon click restore path.

### DIFF
--- a/DockDoor/Utilities/DockObserver.swift
+++ b/DockDoor/Utilities/DockObserver.swift
@@ -803,7 +803,7 @@ final class DockObserver {
                 // This prevents native dock's restore from confusing our logic
                 let needsRestore = restorationNeededAtClickTime || currentlyHasMinimizedWindows
 
-                if needsRestore {
+                if needsRestore, Defaults[.restoreAllMinimizedWindowsOnDockClick] {
                     DebugLogger.log("DockClick", details: "\(appName): restoring (needsRestore=true, minimized=\(currentlyHasMinimizedWindows))")
                     restoreAppWindows(windows: windows, app: app, appName: appName)
                 } else if wasFrontmostOnHover, !windows.isEmpty {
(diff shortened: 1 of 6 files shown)


ejbills (changes requested):
Since this behavior is only intended for the minimize action, it should not show when the user has it configured to hide the application. This will prevent confusion with the feature. @wiskiiii

ejbills (approved):

PR #1380, opened 2026-05-26
fix: position active app indicator across displays

### LINKED ISSUE #1269: [BUG] App indicator on second monitor away from dock
## Bug Description
App indicator on second monitor away from dock

## Steps to Reproduce
1. Move to second larger screen
2. see second dock appear
3. App indicator floating high above dock.

## Expected vs Actual Behavior
App indicator should be attached to dock, under running app

## Environment
- macOS version: Tahoe 26.3.1 (a) 
- App version: DockDoctor 1.36

## Screenshots (optional)

- [x] I have reviewed existing issues to ensure this is not a duplicate

maddada:
Hi there, just fixing a bug I missed when I created this feature.

I was having an issue where when I have two screens that have different resolutions and the Dock moves from one of the screens to the other screen (this happens for example when the dock is at the bottom and you put the cursor at the bottom of the screen that doesn't have the dock). 

Then the active app indicator in the Dock was showing up way offset from the actual Dock itself on the bigger screen. I addressed this with this PR and also tested it and ensured that the indicator shifts as soon as the Dock shifts to the other screen (listening for a notification for this, not using polling). The fix works for left and right side mounted docks, not just bottom docks. 

Closes #1269

## Summary
- Convert Dock accessibility frames to AppKit coordinates before positioning the active app indicator
- Resolve the dock item's actual screen from the converted frame
- Calculate auto sizing from the dock inset on that screen instead of assuming the primary display geometry

Used GPT 5.5 xhigh to resolve this issue but checked the code.

### DIFF
--- a/DockDoor/Utilities/ActiveAppIndicator/ActiveAppIndicatorDockDetection.swift
+++ b/DockDoor/Utilities/ActiveAppIndicator/ActiveAppIndicatorDockDetection.swift
@@ -67,6 +67,18 @@ enum ActiveAppIndicatorDockDetection {
         return CGRect(origin: position, size: size)
     }
 
+    private static func appKitFrame(fromAccessibilityFrame frame: CGRect) -> CGRect {
+        let screen = NSScreen.screenFromQuartzPoint(
+            CGPoint(x: frame.midX, y: frame.midY)
+        )
+        let appKitOrigin = DockObserver.nsPointFromCGPoint(
+            CGPoint(x: frame.minX, y: frame.maxY),
+            forScreen: screen
+        )
+
+        return CGRect(origin: appKitOrigin, size: frame.size)
+    }
+
     /// Calculates indicator height, offset, and length based on dock size and position.
     /// Values derived from testing for dock sizes 36-156.
     /// - Parameters:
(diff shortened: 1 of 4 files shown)


DockDoor/Utilities/ActiveAppIndicator/ActiveAppIndicatorDockDetection.swift
```diff
@@ -67,6 +67,37 @@ enum ActiveAppIndicatorDockDetection {
         return CGRect(origin: position, size: size)
     }
 
+    /// Converts Accessibility's top-left based global frame into the AppKit
+    /// bottom-left based global frame expected by NSWindow positioning.
+    private static func appKitFrame(fromAccessibilityFrame frame: CGRect) -> CGRect? {
```
ejbills: Please refer to the mouse screen / cg converter methods in the codebase - this might be duplicating some logic

Also some logic for determining which screen a point is on is also available somewhere in the codebase - again related to the mouse screen logic

maddada: Done! Sorry for missing this.
Please let me know if there's anything else.

Btw, I noticed that the indicator was moving between screens, so I made it go to opacity 0 when we detect that the dock screen changed, then when it settles on the other screeen (After ~2s) then it gets opacity back.

Issue #1405, opened 2026-06-01
[FR] Show single window for all tabs


chavlji:
## What problem does this solve?
Show only real windows. Do not add an item for each tab inside the window (e.g. Finder tabs).

## Proposed solution
Add a new checkbox in the "Sorting & grouping" section:
"□ Show single window for all tabs"

## Additional context (optional)
Some apps like Finder and Fork offer system of tabs. DockDoor recognizes each tab as a separate window which clutters the experience. There already is an option to show only most recent window of an app, but it:
- In case of Fork it sometimes works incorrectly - it doesn't update which tab  is active and when switching the window with alt+tab it activates some previous tab (instead of just focusing the window)
- In case of Finder user can have multiple windows opened and using this option he won't be able to switch between these windows.

AltTab app has this functionality, but I like DockDoor more, because it's blazingly fast and uncluttered.

- [x] I have reviewed existing issues to ensure this is not a duplicate

It is similar to: https://github.com/ejbills/DockDoor/issues/1034, but not 1:1.

matekoszta:
+1 for this — I'd use it constantly. The core issue is that DockDoor treats native macOS tabs (NSWindow tabbing, used by Terminal, Finder, etc.) as separate windows, because the Accessibility API exposes each tab as its own AXWindow.
The behavior I'm after is exactly what already happens with Chrome/Safari: those browsers use a custom (non-native) tab UI, so the OS reports a single window and DockDoor shows one Alt-Tab entry.

sugarshock:
+1, i tend to have multiple windows of my IDE open (one per project), consisting of multiple tabs per windows. I'd love to be able to activate a "merge tabs of a windows" config somewhere

Maple38:
I need this feature very much — it's quite annoying without it.

We already have a setting to do this on a per-app basis, so it should definitely be doable. Just need a way of inverting the setting so it's a blacklist rather than a whitelist.

Issue #1034, opened 2026-01-14
[BUG] Ghostty tabs are considered different individual windows


ctffarley:
## Bug Description
When using DockDoor and [Ghostty](https://ghostty.org/), the individual tabs within Ghostty are considered windows by DockDoor, in the sense that when you go to the Option+Tab view, each tab shows up instead of just the single Ghostty window. 

I don't know anything about MacOS development, but I know Ghostty experiences issues with certain tiling window managers based on how it handles tabs, see [here](https://ghostty.org/docs/help/macos-tiling-wms). I don't experience this with AltTab so this is technically feasible for this to be solved in DockDoor.

## Steps to Reproduce
1. Open Ghostty
2. Open at least one additional tab
3. Use the DockDoor Option+Tab view and see the Ghostty tabs showing up as windows

## Expected vs Actual Behavior
I would expect to only see the singular Ghostty window I have open show up. Instead, the individual tabs show up as their own windows.

## Environment
- macOS version: Sonoma 14.2.1
- App version: 1.30.1

## Screenshots (optional)

https://github.com/user-attachments/assets/89d7dbfe-c6f4-4340-8edf-d39b773ab801

- [x] I have reviewed existing issues to ensure this is not a duplicate

ejbills:
<img width="1000" height="666" alt="Image" src="https://github.com/user-attachments/assets/42544a7e-52cd-412b-bc1a-16733ef3de8c" />

Can you check if it is tied to this setting in the macOS settings? If it is, this is a duplicate of #631.

ctffarley:
Regardless of the option, it seems to exhibit this behavior. The weird thing is that the number of tabs in Ghostty seems to never equal the number of tabs that show up in DockDoor, unlike that other issue you linked. It seems to max out at 3. For a second it seemed like "In Full Screen" was working but it also had the issue of switching to the latest tab upon using Option+Tab, and then when I added a new tab that tab showed up in the window picker.

maddada:
Maybe this option will help here?

<img width="1116" height="990" alt="Image" src="https://github.com/user-attachments/assets/83047623-2a94-46ca-80a1-baf2d425bc94" />

ctffarley:
@maddada thanks, this does seem to have worked, with the slightest caveat:

[...]

PR #1445, opened 2026-06-17
feat: collapse native macOS tabs into a single window preview

### LINKED ISSUE #1434: Feature: Option to Collapse Tabs into Single Preview for Tabbed Apps (e.g. Finder)
## Problem
Currently, when using DockDoor with apps that support tabs (such as Finder), each tab within a single window is detected and displayed as a separate preview card in Dock/Dock hover and window switcher views. This leads to visual clutter, especially for users with several tabs open in one Finder window, as each Finder tab appears as an independent window in DockDoor’s UI. For most users, the intuitive expectation is that a window with multiple tabs should appear as a single item, mirroring macOS’s own app switcher (Cmd+Tab) and window management behavior.

## Proposed Enhancement
Add an option (either global or app-specific – Finder is most important) to *collapse or merge all tabs from the same physical/app window into a single preview card* in DockDoor. Only the currently visible (frontmost/selected) tab should be shown as the preview.

## Example
If Finder has one window with 3 tabs, DockDoor should only show 1 preview for that window (the active tab), not 3.

## Additional context
Attached are images demonstrating the current behavior for Finder:

Image

Image

romankhadka:
## What

Adds an opt-in setting that collapses native macOS window-tab groups so a tabbed window shows up as a single preview instead of one preview per tab.

Closes #1434. Also addresses #1034 (Ghostty tabs counted as separate windows) and #1405.

## Why

Apps that use native macOS window tabbing (Ghostty, Finder, Terminal, Safari, and others) model each tab as a distinct `NSWindow`. The Accessibility API reports each of those windows separately, and window discovery had no tab-group awareness, so one tabbed window surfaced as one preview per tab.

There is no Accessibility attribute that exposes tab-group membership for arbitrary apps, so membership is inferred from a reliable side effect: the windows in one visible tab group are stacked at the exact same screen frame. Windows that share a process and an integer-rounded frame are treated as one group.

## How

- New `NativeTabGrouping` enum holds the grouping logic as a pure function over a minimal `Candidate` value type (id, pid, frame, recency, groupable), so it is unit-testable without Accessibility or AppKit types.
- `WindowUtil.collapseNativeTabsIfNeeded(_:)` adapts `WindowInfo` to `Candidate` and applies the grouping at both discovery choke points: the per-app path (`getActiveWindowsUpdatingCache`, used by dock previews and Cmd+Tab) and the all-apps switcher path (`getAllWindowsOfAllApps`). Because the group key includes the process id, passing the multi-app switcher list never merges windows from different apps.
- The representative kept per group is the most recently accessed window (typically the active tab); ties break on the larger window id for determinism.
- Minimized, hidden, windowless, and zero-sized windows are never collapsed.
- New toggle "Collapse native tabs into a single window" under Dock Previews > Window Display, plus a settings-search entry. Defaults to off.

## Trade-off

Collapsing a tab group also collapses genuinely separate windows of the same app that happen to share a frame. The active-app window-switcher hotkey still cycles individual windows, which is why this ships as an opt-in default-off setting.

## Testing

- `NativeTabGroupingTests` covers same-frame collapsing, distinct frames, cross-process isolation, non-groupable pass-through, subpixel rounding, deterministic tie-break, and empty input.
- Full app target and test target build clean; `xcodebuild test` passes locally (Xcode 26.5, macOS arm64).

### DIFF
--- a/DockDoor.xcodeproj/project.pbxproj
+++ b/DockDoor.xcodeproj/project.pbxproj
@@ -60,6 +60,7 @@
 		AA11001122334455AA110022 /* DockLocker.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA11001122334455AA110011 /* DockLocker.swift */; };
 		AA11001122334455AA110044 /* DockLockingSettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA11001122334455AA110033 /* DockLockingSettingsView.swift */; };
 		AA11001122334455AA110066 /* DockLockerGeometryTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA11001122334455AA110055 /* DockLockerGeometryTests.swift */; };
+		555110B7F7B99C7BD510DBC7 /* NativeTabGroupingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D00D4B366E3CF5338182E065 /* NativeTabGroupingTests.swift */; };
 		B376D66B467C4587915220BC80D3B9CB /* ScriptCommands.swift in Sources */ = {isa = PBXBuildFile; fileRef = F098F8107F464776B5C18294427CAECD /* ScriptCommands.swift */; };
 		B3C9F20484F14266B941AFB170B66530 /* MediaRemoteService.swift in Sources */ = {isa = PBXBuildFile; fileRef = AC298B81953848F481EE849906E0AE77 /* MediaRemoteService.swift */; };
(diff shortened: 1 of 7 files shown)


ejbills (approved):

uday-appdirect:
damn, I just came and reported a [feature request.](https://github.com/ejbills/DockDoor/issues/1465) seems like its already merged

Any ETA for next release?

ejbills:
@uday-appdirect you can try the latest beta and report issues

Issue #1524, opened 2026-07-30
[BUG] MediaRemoteAdapter perl helper leaks ~8MB per media event (3.3GB in 3h); enableMediaWidget=false doesn't stop it


sri-c9:
## Bug Description

The `MediaRemoteAdapter` perl helper leaks roughly 8 MB per media event and never releases it. On my machine it reached **3.29 GB after 3 hours** of normal use.

Two separate problems, the second making the first much worse:

**1. The helper leaks.** `footprint` on the helper process:

```
perl [99747]: 64-bit    Footprint: 3367 MB (16384 bytes per page)

  Dirty      Clean  Reclaimable    Regions    Category
3351 MB        0 B          0 B        423    MALLOC_LARGE
7056 KB        0 B          0 B         20    MALLOC metadata
...
3367 MB     128 KB          0 B       4612    TOTAL
```

All 423 leaked regions are **exactly 8112 KB** — a fixed-size buffer allocated per media event and never freed, not variable-size payload data. The buffers are never read after being written, so macOS compresses all of them:

```
Writable regions: Total=3.3G written=3.3G(99%) resident=2931K(0%) swapped_out=3.3G(99%)
```

That's why this is easy to miss: `ps` RSS reports ~10 MB. Only Activity Monitor's Memory column (`phys_footprint`) shows the real 3.29 GB. Memory is never reclaimed when playback stops — it only ever grows for as long as DockDoor runs.

**2. `enableMediaWidget` doesn't gate the listener.** `AppDelegate.swift:62` calls:

```swift
MediaRemoteService.shared.activate()
```

unconditionally, with no `Defaults[.enableMediaWidget]` check. `MediaRemoteService.activate()` only guards on `!isActive` before calling `controller.startListening()`. I verified this: with `enableMediaWidget = false` set and DockDoor relaunched, the perl helper still spawns. So a user who disables the media widget still pays its full memory cost with no way to opt out short of quitting DockDoor.

## Steps to Reproduce

1. Run DockDoor with a media app (Apple Music) playing, changing tracks normally for a few hours.
2. Sort Activity Monitor by Memory — a `perl` process owned by DockDoor climbs into the GBs.
[...]

PR #1526, opened 2026-07-31
fix: gate media listener on enableMediaWidget setting

inequitas:
Addresses the second half of #1524 — the part that lives in DockDoor rather than in the vendored adapter.

## Problem

`MediaRemoteService.activate()` was called unconditionally from `AppDelegate.swift:62` and had no `Defaults[.enableMediaWidget]` check of its own, so `controller.startListening()` always ran. That spawns a `run.pl` helper which lives for the rest of the session, so a user who had turned the media widget off still paid its full memory cost with no way to opt out short of quitting DockDoor.

`deactivate()` already existed but had no callers, so toggling the setting at runtime did nothing either.

## Change

Two small things in `MediaRemoteService`:

- `activate()` now returns early when `enableMediaWidget` is off.
- A `Defaults` observer starts/stops the listener when the setting is toggled, so turning the widget off takes effect immediately rather than at the next launch.

The guard sits in `activate()` rather than at the `AppDelegate` call site so the precondition travels with the service — `activate()` is the only thing that can spawn the helper, so it seemed like the right place for it to enforce its own precondition.

## Verification

macOS 26.5 (Apple Silicon), Xcode 26.6, counting `perl` children of the running app with the setting written to the debug domain:

| build | `enableMediaWidget` | perl helpers |
|---|---|---|
| before | `false` | **1** ← the bug |
| after | `false` | **0** |
| after | `true` | 1 (unchanged) |

I also confirmed the before/after by stashing the change and rebuilding, rather than reasoning about it — the helper genuinely spawns on `main` today and genuinely does not with this applied.

- `xcodebuild` Debug and Release both succeed, no new warnings (the two existing `SpaceWindowCacheManager.swift` lock/unlock warnings are untouched)
- `swift run -c release --package-path BuildTools swiftformat --lint` reports `0/1 files require formatting`

## Scope

This does **not** address the leak itself, which is the more interesting half of #1524. The reporter's comparison against boringNotch — same library, `stream` vs `loop` mode, framework directory vs binary path, 11 MB over 3 days vs 3.3 GB over 3 hours — points at the invocation path rather than at DockDoor's Swift code, and that's a separate change I didn't want to bundle in here.

What this does mean is that users who don't want the media widget can now actually avoid the helper entirely, which is a complete workaround for them in the meantime.

Happy to adjust the placement of the guard if you'd rather it sat in `AppDelegate` alongside the other `Defaults` checks there.

### DIFF
--- a/DockDoor/Utilities/Media/MediaRemoteService.swift
+++ b/DockDoor/Utilities/Media/MediaRemoteService.swift
@@ -1,5 +1,6 @@
 import AppKit
 import Combine
+import Defaults
 import Foundation
 import MediaRemoteAdapter
 
@@ -39,6 +40,8 @@ final class MediaRemoteService: ObservableObject {
     private var playbackIntent: (isPlaying: Bool, date: Date)?
 
     private var resyncWork: DispatchWorkItem?
+
+    private var mediaWidgetObserver: Defaults.Observation?
 
     var interpolatedElapsedTime: TimeInterval {
         if let seek = seekBase, seek.rate == 0 || isWithinSeekWindow {
@@ -80,6 +83,7 @@ final class MediaRemoteService: ObservableObject {
 
     private init() {
         setupController()
+        observeMediaWidgetSetting()
     }
 
     private func setupController() {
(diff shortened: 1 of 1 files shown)


ejbills:
100% AI generated

PR #1545, opened 2026-08-10
feat(cmd-tab): restore minimized/hidden windows when switching (AltTab-style)

linmingyu8:
## Summary

When using Cmd+Tab with DockDoor's enhancements, switching to an app whose target window is **minimized or hidden** did nothing — the window stayed in the Dock. This mirrors AltTab / Windows Alt+Tab, which restore the window to the front.

## Root cause

`WindowInfo.bringToFront()` relies on `_SLPSSetFrontProcessWithOptions` / `makeKeyWindow` / `kAXRaiseAction`, which **silently fail on minimized/hidden windows**: the app activates, but the window never comes forward.

## Changes

- **`WindowInfo.swift`** — added `isPutAway` (`isMinimized || isHidden`) and `restoreFromPutAwayState()` (unhide the app, then deminiaturize the window — mirroring AltTab's `Window.focus()`). `bringToFront()` now restores put-away windows **first**, so every focus path is covered (Cmd+Tab selection, Window Switcher, preview click, drag). `toggleMinimize()` reuses the same logic.
- **`DockObserver+CmdTab.swift`** — app-level fallback: when Cmd is released without an explicit window selection and the target app's windows are **all** minimized/hidden, restore the most recently used one (AltTab/TabLift behavior).
- **`KeybindHelper.swift`** — wires the fallback into the Cmd-release path.
- **`consts.swift`** + **`CmdTabSettingsView.swift`** + **`SettingsSearchCatalog.swift`** — new setting `restoreMinimizedWindowsOnCmdTab` (default **on**) with a UI toggle and a settings-search entry.

## Behavior

- Selecting a minimized/hidden window via Cmd+Tab now brings it forward.
- App-level fallback restores a fully-minimized app when Cmd is released without an explicit pick.
- The new toggle lets users keep the old behavior if they prefer.

## Notes

- Default on; disable `Restore minimized/hidden windows on switch` in Cmd+Tab settings to revert.
- Needs Xcode to build (no CI change). Please compile-test on macOS before merging.

### DIFF
--- a/DockDoor/Utilities/DockObserver+CmdTab.swift
+++ b/DockDoor/Utilities/DockObserver+CmdTab.swift
@@ -17,6 +17,7 @@ extension DockObserver {
         cmdTabObserver = nil
         stopCmdTabPolling()
         DockObserver.isCmdTabSwitcherActive = false
+        DockObserver.lastCmdTabTargetApp = nil
     }
 
     // MARK: - On-Demand Polling (Event-Driven)
@@ -118,6 +119,7 @@ extension DockObserver {
         }
 
         let resolvedApp = selectedItem.app
+        DockObserver.lastCmdTabTargetApp = resolvedApp
         let appName = resolvedApp?.localizedName ?? selectedItem.title ?? "Unknown"
         let bundleId = resolvedApp?.bundleIdentifier ?? selectedItem.bundleId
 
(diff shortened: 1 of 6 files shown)


ejbills:
Use the PR template

PR #1590, opened 2026-09-04
fix: proper handling of the case when the same app has multiple instances running and multiple dock icons

Zoinen:
## Describe your changes
When there's more than one instance of the same app running, and it's being shown as multiple icons in dock, there were multiple issues in the current handling of this case:
1. active app indicator didn't properly detect which instance was active, and it was always highlighting the first one
2. enabling "click to minimize" on dock icons was incorrectly treating switching between different instances as clicks on the same entity so it was hiding the previous one when clicked on the next one
3. active app indicator was showing slide animation when switching between app instances instead of moving indicator instantly to a newly activated entity and showing proper animation (like it does when switching between any other apps).

My changes map duplicate Accessibility Dock items to running applications using launch order and persistent Dock item count instead of relying on NSRunningApplication array order, also I'm using the process identity for active indicator transitions. Added tests for persistent and transient instance ordering.

- [+] I have performed a self-review of my code and understand every line
- [+] If this change affects core functionality, I have added a description highlighting the changes
- [+] I have followed conventional commit guidelines
- [+] I have tested this PR on my own machine

## AI Assistance Disclosure
- [+] Yes - describe below how AI was used and confirm you reviewed/tested the output
Yes, I used Codex with GPT 5.6 Luna Max, it helped me find the exact code to fix the bug I found, I did tested thoroughly on my machine and it works fine


Zoinen:
found one more bug, don't merge just yet

Zoinen:
The additional closed-instance bug is fixed in 7ed3f90. I verified both surviving-tile arrangements: the indicator stays with the running process, and clicking an inactive duplicate no longer minimizes or otherwise targets the survivor. The PR description now includes this case.

PR #1599, opened 2026-09-07
fix: dismiss stale full-size hover previews

kiteretsu903:
## Describe your changes

I've run into this bug for a long time and kept hoping an update would fix it. It seemed random, but I finally reproduced it reliably with a debugging script.

To reproduce:

1. Select “Present a full size preview of the window” with a 0.3-second delay.
2. Hover over Calculator in the Dock, then move onto its thumbnail.
3. Press Escape within 0.1–0.25 seconds, keeping the pointer still.
4. The thumbnail closes, but the full-size preview appears and stays stuck—even after another Escape.

This fix cancels stale hover requests and lets Escape dismiss the full-size preview independently.

Validation: 38 tests passed, including six local regression tests. Native input replay passed 12 timing and dismissal scenarios, with additional visual checks. I haven't included the test script or local regression tests in this PR; let me know if you'd like them.

None linked.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex helped reproduce the bug, implement the fix, run tests, and draft this description. I reviewed the code, understand the changes, and tested them on my own machine.

## Core Functionality Changes

Delayed full-size previews require an active hover on a visible thumbnail panel. Dismissal invalidates pending requests; Escape also closes a remaining full-size preview.

### DIFF
--- a/DockDoor/Views/Hover Window/Shared Components/SharedPreviewWindowCoordinator.swift
+++ b/DockDoor/Views/Hover Window/Shared Components/SharedPreviewWindowCoordinator.swift
@@ -24,6 +24,7 @@ final class SharedPreviewWindowCoordinator: NSPanel {
     var mouseIsWithinPreviewWindow: Bool = false
     private var onWindowTap: (() -> Void)?
     private var fullPreviewWindow: NSPanel?
+    private var activeFullPreviewHoverID: UUID?
     private var pendingShowWorkItem: DispatchWorkItem?
 
     var windowSize: CGSize = getWindowSize()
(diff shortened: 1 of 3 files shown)


DockDoor/Views/Hover Window/Shared Components/SharedPreviewWindowCoordinator.swift
```diff
@@ -1102,6 +1123,9 @@ final class SharedPreviewWindowCoordinator: NSPanel {
             }
 
             Task { @MainActor [weak self] in
+                if centeredHoverWindowState == .fullWindowPreview, self?.isFullPreviewHoverActive(fullPreviewHoverID) != true {
```
ejbills, line 1126: Why do we bail here?

kiteretsu903: > Just making sure we don’t show a preview the user already dismissed. The DispatchWorkItem creates a separate Task, and canceling pendingShowWorkItem won’t cancel that Task once it’s queued. So if you hit esc before the Task runs, without this check the preview could pop back up.

